PR #17, opened 2020-01-28
Generate sphinx doc from docstring

### LINKED ISSUE #15: Generate documentation from docstring
Currently, sphinx only contains a handful of notes about RBC. It would be good if one could set up it to generate documentation from docstrings.

guilhermeleobas:
Fixes #15


doc/rbc/api.rst
```diff
@@ -57,9 +57,9 @@ These methods are in the ``rbc.omnisci_array`` module namespace
 RemoteNumba
```
pearu: Let's drop remotenumba for simplicity.

guilhermeleobas:
I think this is ready for merging. Any other addition can be done in a different PR. What do you think?

PR #21, opened 2020-01-29
Remove duplicate initialize_llvm function

guilhermeleobas:
It looks like that `initialize_llvm` was defined twice [`irtools.py`](https://github.com/xnd-project/rbc/blob/446514cff60e8ee2ab43228fc506128a54f0ca47/rbc/irtools.py#L11-L14) and [`remotenumba.py`](https://github.com/xnd-project/rbc/blob/446514cff60e8ee2ab43228fc506128a54f0ca47/rbc/remotenumba.py#L16-L19) but it's only used on [`remotenumba.py`](https://github.com/xnd-project/rbc/blob/446514cff60e8ee2ab43228fc506128a54f0ca47/rbc/remotenumba.py#L22). This PR deletes the former.


guilhermeleobas:
Move everything from `remotenumba.py` to `utils/` or just the `initialize_llvm` method?

pearu:
Move `remotenumba.py` to `utils/`.

guilhermeleobas:
Done.

PR #41, opened 2020-03-04
Add NumPy types mapping

### LINKED ISSUE #40: Add numpy types on type system
The following code fails:

```python
@rjit('double(double, double)')
    def logaddexp(x, y):
        return np.logaddexp(x, y)

x = np.log(1e-50)
y = np.log(2.5e-50)
logaddexp(x, y)
```

```python
...
rbc/tests/test_numpy_rjit.py:50:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
rbc/remotejit.py:261: in __call__
    atypes = tuple(Type.fromvalue(a, target_info) for a in arguments)
rbc/remotejit.py:261: in <genexpr>
    atypes = tuple(Type.fromvalue(a, target_info) for a in arguments)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

cls = <class 'rbc.typesystem.Type'>, obj = -115.12925464970229, target_info = <rbc.targetinfo.TargetInfo object at 0x10c5766d0>

    @classmethod
    def fromvalue(cls, obj, target_info):
        """Return Type instance that corresponds to given Python value.
        """
        n = _python_imap.get(type(obj))
        print(type(obj))
        if n is not None:
            return cls.fromstring(n, target_info)
>       raise NotImplementedError('%s.fromvalue(%r)'
                                  % (cls.__name__, obj))
E       NotImplementedError: Type.fromvalue(-115.12925464970229)

rbc/typesystem.py:751: NotImplementedError
```

`x` and `y` have types `numpy.float64`.

guilhermeleobas:
Fixes #40.

Tests that pass a numpy array fails:
```python
@rjit('float[](float[], float[])')
def logaddexp(x, y):
    return np.logaddexp(x, y)

x = [1, 2, 3, 4, 5]
y = [5, 4, 3, 2, 1]
z = [6, 6, 6, 6, 6]
for dt in ['f', 'd', 'g']:
    logxf = np.log(np.array(x, dtype=dt))
    logyf = np.log(np.array(y, dtype=dt))
    logzf = np.log(np.array(z, dtype=dt))
    assert(np.allclose(logaddexp(logxf, logyf), logzf))
```

I will let array support for a future pull request and focus only on this in this one.

Still missing:
- [x] Fix argument parsing when the array is n-dimensional. For instance, `void(float[:, :])` is parsed as `void` -> `float[:`, `:]`, rather than `void` -> `float[:, :]`.
- [ ] ~`_demangle` only 1d-arrays signatures. Fix it to return correct signatures for any ndarray array.~ Left to a future pull request since this will only be used with arrays.


guilhermeleobas:
Ready for a review

rbc/typesystem.py
```diff
@@ -680,6 +681,17 @@ def fromstring(cls, s, target_info):
...
+    def fromnumpy(cls, t, target_info):
+        """Return new Type instance from numpy type object.
+        """
+        if np is None:
+            raise RuntimeError('importing numpy failed: %s' % (np_NA_message))
+
+        n = _numpy_imap.get(t)
+        if n is not None:
+            return cls.fromstring(n, target_info)
+
```
pearu: Insert `raise NotImplementedError(repr(t))` as otherwise `None` is returned.

rbc/typesystem.py
```diff
@@ -212,15 +212,16 @@ def topython(self):
...
 
 # numpy mapping
+_numpy_imap = {}
 if np is not None:
-    _numpy_imap = {np.float32: 'float32', np.double: 'float64',
-                   np.longdouble: 'float128'}
-else:
-    _numpy_imap = {}
+
+    for k, v in np.typeDict.items():
```
pearu: Use `for v in set(numpy.typeDict.values()):`

rbc/tests/test_typesystem.py
```diff
@@ -341,6 +348,44 @@ def fromnumba(t):
...
+    assert fromnumpy(np.object_) == fromstr('object')
+    assert fromnumpy(np.str_) == fromstr('str')
+    assert fromnumpy(np.timedelta64) == fromstr('timedelta64')
+    assert fromnumpy(np.uint8) == fromstr('uint8')
+    assert fromnumpy(np.uint16) == fromstr('uint16')
+    assert fromnumpy(np.uint32) == fromstr('uint32')
+    assert fromnumpy(np.uint64) == fromstr('uint64')
+    assert fromnumpy(np.ulonglong) == fromstr('uint64')
+
+    if not win32:
```
pearu: Non-win32 platforms may also miss `numpy.float128` (e.g. 32-bit linux).
Suggest using `if hasattr(np, 'float128')`.

rbc/tests/test_typesystem.py
```diff
@@ -3,11 +3,18 @@
...
+except ImportError:
+    np = None
+
+import sys
 import pytest
 from rbc.typesystem import Type
 from rbc.utils import get_datamodel
 from rbc.targetinfo import TargetInfo
 
+win32 = sys.platform == 'win32'
```
pearu: not needed, see below

guilhermeleobas:
@pearu, what would be the correct signature for a multidimensional array? Let's say I have a 3d float32 array, the signature would be `float32[][][]`?

pearu:
Numba uses `float32[:, :, :]` convention to create a 3d array type.

guilhermeleobas:
Current status: I am trying to pass a ctypes structure for a numba jitted function that was compiled for an [ArrayModel](https://github.com/numba/numba/blob/13ece9b97e6f01f750e870347f231282325f60c3/numba/core/datamodel/models.py#L857) but it is not working. I tried to get this working with a [minimal example](https://llvmlite.readthedocs.io/en/latest/user-guide/binding/examples.html) and it seems that values are being treated as 64 bits, even though they are declared as 32 bits.

I will remove related code to numpy array from this PR to get it merged.

guilhermeleobas:
@pearu, can you review this PR?

guilhermeleobas:
Limited the scope of this PR to just the NumPy types support.

PR #44, opened 2020-03-05
Add floating-point equivalent functions to libm whitelist

### LINKED ISSUE #43: libm whitelist should contain the floating-point equivalents
The code below breaks with an exception saying that `log1pf` is not defined. This is just a matter of whitelisting libm functions with an `f` suffix: `log1pf, asinhf, cosf`. For functions which receive a long double, we must use the `l` suffix.

```python
def test_foo(rjit):
    @rjit('float(float, float)')
    def logaddexp(x, y):
        return np.logaddexp(x, y)

    x = [1, 2, 3, 4, 5]
    y = [5, 4, 3, 2, 1]
    z = [6, 6, 6, 6, 6]
    for dt in ['f', 'd', 'g']:
        logxf = np.log(np.array(x[0], dtype=dt))
        logyf = np.log(np.array(y[0], dtype=dt))
        logzf = np.log(np.array(z[0], dtype=dt))
        assert(np.allclose(logaddexp(logxf, logyf), logzf)) 
```

guilhermeleobas:
Fix #43

### DIFF
--- a/rbc/irtools.py
+++ b/rbc/irtools.py
@@ -19,9 +19,12 @@ fp_funcs = ['frexp', 'ldexp', 'modf', 'scalbn', 'scalbln', 'nextafter',
             'nexttoward']
 classification_funcs = ['fpclassify', 'isfinite', 'isinf', 'isnan',
                         'isnormal', 'signbit']
-libm_funcs = [*exp_funcs, *power_funcs, *trigonometric_funcs,
-              *hyperbolic_funcs, *nearest_funcs, *fp_funcs,
-              *classification_funcs]
+def lf(l):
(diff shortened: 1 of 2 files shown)


rbc/irtools.py
```diff
@@ -8,6 +8,11 @@
 from .targetinfo import TargetInfo
 from .npy_mathimpl import *  # noqa: F403, F401
 
+
+def lf(l):
+    return l + [e + 'f' for e in l] + [e + 'l' for e in l]
```
pearu: Make `lf` private: rename it to `_lf`.
Suggest not using `l` as variable name (as it looks like `1`). Use `lst`, for instance.

PR #62, opened 2020-04-15
Initial support for Array(dtype, size)

guilhermeleobas:
Still a work in progress. 

This PR adds the ability to instantiate an Omnisci array inside a jitted code.

I will keep the list below as a To-Do list:
- [x] Find a way to get `TargetInfo`  inside `type_omnisci_array` or `array_type_converter`
- [x] Formalize the signature used to construct arrays. Are we down for using the type as a string? i. e. `int32[]`. 
- [x] Find a way to get the length in bytes of the type `T`. OmniSci uses a `reinterpreter_cast<T>` but we don't have this fancy machinery at hand. The typesystem can be used in this case: `typesystem.Type.fromstring(int32).bits`
- [ ] Check how memory is released
- [x] Change tests to eval an expression without a `table` statement.
- [x] Change the order of arguments inside the constructor
- [x] Remove square brackets from the dtype when constructing a new array
- [x] Add a method to access the `is_null` property
- [x] Support other types on `dtype`. i.e. `Array(size, dtype=types.int32)` or `Array(size, dtype=np.int32)`

### DIFF
--- a/rbc/omnisci_array.py
+++ b/rbc/omnisci_array.py
@@ -34,10 +34,8 @@ class Array(object):
     pass
 
 
-@extending.lower_builtin(Array, types.Integer, types.StringLiteral)
-def omnisci_array_constructor(context, builder, signature, args):
+def omnisci_array_constructor(context, builder, signature, args, elsize):
     pyapi = context.get_python_api(builder)
-    targetinfo = TargetInfo.host()
 
     # integer types used
     i8 = ir.IntType(8)
@@ -45,8 +43,6 @@ def omnisci_array_constructor(context, builder, signature, args):
 
     # grab args
     sz, _ = args
-    dtype = signature.args[1].literal_value
-    elsize = typesystem.Type.fromstring(dtype, targetinfo)  # element size
     elsize_ir = context.get_value_type(elsize.tonumba())  # get the ir type
 
     # fill 'sz' and 'is_null'
@@ -64,11 +60,38 @@ def omnisci_array_constructor(context, builder, signature, args):
     return fa._getpointer()
 
 
+@extending.lower_builtin(Array, types.Integer, types.StringLiteral)
+def omnisci_array_constructor_string_literal(context, builder, signature, args):
+    targetinfo = TargetInfo.host()
+
+    dtype = signature.args[1].literal_value
+    elsize = typesystem.Type.fromstring(dtype, targetinfo)  # element size
+
+    return omnisci_array_constructor(context, builder, signature, args, elsize)
+
+
+@extending.lower_builtin(Array, types.Integer, types.NumberClass)
+def omnisci_array_constructor_numba_type(context, builder, signature, args):
+    targetinfo = TargetInfo.host()
+
+    it = signature.args[1].instance_type
+    elsize = typesystem.Type.fromnumba(it, targetinfo)
+
+    return omnisci_array_constructor(context, builder, signature, args, elsize)
+
+
 @extending.type_callable(Array)
 def type_omnisci_array(context):
     def typer(size, dtype):
         targetinfo = TargetInfo.host()
-        conv = array_type_converter(targetinfo, dtype.literal_value + '[]')
+
+        if isinstance(dtype, types.NumberClass):
+            it = dtype.instance_type
+            typ = typesystem.Type.fromnumba(it, targetinfo).tostring() + '[]'
+        elif isinstance(dtype, types.StringLiteral):
+            typ = dtype.literal_value + '[]'
+
+        conv = array_type_converter(targetinfo, typ)
         return conv._params['tonumba']
     return typer
 
--- a/rbc/omnisci_array.py
+++ b/rbc/omnisci_array.py
@@ -4,6 +4,7 @@ from llvmlite import ir
 import numpy as np
 from . import typesystem
 from .utils import get_version
+from .targetinfo import TargetInfo
 if get_version('numba') >= (0, 49):
     from numba.core import datamodel, cgutils, extending, types
 else:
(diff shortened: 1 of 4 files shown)


rbc/omnisci_array.py
```diff
@@ -29,6 +30,28 @@ def key(self):
...
+    _, sz = args
+    fa = cgutils.create_struct_proxy(typ)(context, builder) 
+    fa.sz = sz
+    return fa._getpointer()
+
+
+@extending.type_callable(Array)
+def type_omnisci_array(context):
+    def typer(dtype, size):
+        targetinfo = TargetInfo.host()
```
guilhermeleobas, line 49: Not sure if instantiating the `TargetInfo` here is 100% correct.

pearu: no, it is not. The targetinfo should contain server info. We'll need to find a way to get the server TargetInfo from RemoteOmnisci instance to here..

pearu: For instance, via `array_type_converter` function in omnisci_array.py .

guilhermeleobas:
Hi @pearu, can you review this PR? Some parts are a bit hacky and they will be addressed before this PR gets merged.

rbc/omnisci_array.py
```diff
@@ -29,6 +30,56 @@ def key(self):
...
+    pass
+
+
+# XXX there should be a better way to find element size
+bytes_map = {
+    'int8': 1,
+    'int16': 2,
+    'int32': 4,
+    'int64': 8
+}
```
guilhermeleobas, line 43: This is "necessary" for the final bitcast instruction. There's probably a better way to detect the type length using the typesystem.

rbc/tests/test_omnisci_array.py
```diff
@@ -280,3 +280,77 @@ def array_setitem_sum(b, c):
...
+    def array_len(size):
+        a = Array('double[]', size)
+        return len(a)
+
+    query = (
+        'select array_len(3) from {omnisci.table_name}'
+        .format(**locals()))
+    _, result = omnisci.sql_execute(query)
+
+    print(list(result))
```
guilhermeleobas, line 300: You can ignore the tests for now.

rbc/omnisci_array.py
```diff
@@ -29,6 +30,56 @@ def key(self):
...
+    pyapi = context.get_python_api(builder)
+
+    # integer types used
+    i8 = ir.IntType(8)
+    i64 = ir.IntType(64)
+
+    # grab args
+    _, sz = args
+    dtype = sig.args[0].literal_value.strip('[]')
+    elsize = bytes_map[dtype]  # element size in bytes
```
pearu, line 57: Use:
```
elsize = sig.args[0].literal_type.bitwidth // 8
```
or similar (I don's see why to strip `[]` atm).

rbc/omnisci_array.py
```diff
@@ -29,6 +30,56 @@ def key(self):
...
+
+    # fill 'sz' and 'is_null'
+    typ = sig.return_type.dtype
+    fa = cgutils.create_struct_proxy(typ)(context, builder)
+    fa.sz = builder.zext(sz, i64)  # zero-extend the size to i64
+    fa.is_null = ir.Constant(i8, 0)
```
pearu, line 63: Use:
```
fa.is_null = i8(0)
```

rbc/tests/test_omnisci_array.py
```diff
@@ -280,3 +280,77 @@ def array_setitem_sum(b, c):
...
+    omnisci.reset()
+
+    from rbc.omnisci_array import Array
+
+    @omnisci('int32(int32)')
+    def array_ptr(size):
+        a = Array('int32[]', size)
+        for i in range(size):
+            a[i] = i
+        return a[2]
```
pearu, line 313: Just wondering, how or where the memory of `a` is released?

rbc/omnisci_array.py
```diff
@@ -29,6 +30,56 @@ def key(self):
...
+
+    # fill 'ptr' with the return value of 'allocate_varlen_buffer'
+    fnty = ir.FunctionType(i8.as_pointer(), [i64, i64])
+    fn = pyapi._get_function(fnty, name="allocate_varlen_buffer")
+    call = builder.call(fn, [fa.sz, ir.Constant(i64, elsize)])
```
pearu, line 68: Use `i64(elsize)` instead of `ir.Constant(...)`.

rbc/omnisci_array.py
pearu: What is `context` here? If it is RemoteCPUContext instance then `context.target_info` is the solution.

guilhermeleobas:
With this last commit, I've addressed almost all items on the to-do list, @pearu. Can you review this PR when you have some free cycles to spare?

rbc/omnisci_array.py
```diff
@@ -29,11 +30,104 @@ def key(self):
...
+
+    return fa._getpointer()
+
+
+@extending.lower_builtin(Array, types.Integer, types.StringLiteral)
+def omnisci_array_constructor_string_literal(context, builder, sig, args):
+    targetinfo = TargetInfo.host()
+
+    dtype = sig.args[1].literal_value
+    elsize = typesystem.Type.fromstring(dtype, targetinfo)  # element size
```
pearu, line 68: Use
```
elsize = typesystem.Type.fromstring(dtype, context.target_info)
```
Consider renaming `elsize` to `eltype` or similar.

rbc/omnisci_array.py
```diff
@@ -29,11 +30,104 @@ def key(self):
...
+
+    return omnisci_array_constructor(context, builder, sig, args, elsize)
+
+
+@extending.lower_builtin(Array, types.Integer, types.NumberClass)
+def omnisci_array_constructor_numba_type(context, builder, sig, args):
+    targetinfo = TargetInfo.host()
+
+    it = sig.args[1].instance_type
+    elsize = typesystem.Type.fromnumba(it, targetinfo)
```
pearu, line 78: Use
```
typesystem.Type.fromnumba(it, context.target_info)
```

rbc/omnisci_array.py
pearu: Use
```
target_info = context.target_info
```

rbc/omnisci_array.py
```diff
@@ -4,6 +4,7 @@
 import numpy as np
 from . import typesystem
 from .utils import get_version
+from .targetinfo import TargetInfo
```
pearu, line 7: Remove this import statement as not needed.

guilhermeleobas:
From what I got, we have two cases from managing memory:
1. *Create a temporary array*: Need to free the memory ourselves.
2. *Create an array and return it*: Register the memory to be freed by the OmniSci runtime.

In the first case, we need to keep track of all arrays created and insert code to `free` the ones that will not be returned. This must be done inside basic blocks with `return` stmt.

In the second case, OmniSci has a method called `register_buffer_with_executor_rsm` to register buffers to be freed.

guilhermeleobas:
I came up with the following scenarios

```python

def case1(sz, typ):
    """
    Only register the array to be freed by the omnisci runtime engine
    """
    return Array(sz, typ)

def case2(sz, typ):
    """
    Alloc the memory but access only a property.
    Register the memory to be freed by omnisci
    """
    a = Array(sz, typ)
    return len(a) # or is_null(a)

def case3(sz, typ, v):
    """
    Use the memory to compute something but don't return it
    """
    a = Array(sz, typ)
    a.fill(v)
    return np.sum(a)

def case4(sz, typ):
    """
    Create an alias.
    """
    a = Array(sz, typ)
    b = a # alias
    b.fill(5)
  return b

def case5(sz, typ):
    """
    Free one and return the other
    """
    a = Array(sz, typ)
    a.fill(3)
    b = Array(sz, typ)
    b.fill(5)
    if sz % 2 == 0:
        return a
    else:
        return b
```

guilhermeleobas:
Pearu, PR is ready for merge. I will tackle the memory management problem on PR https://github.com/xnd-project/rbc/pull/68.

PR #68, opened 2020-05-01
Omnisci Memory management

guilhermeleobas:
This pull request tackles the problem fo memory management for omnisci arrays.

Missing
- [x] Only change the pipeline if `RemoteOmnisci` is used.
- [x] Do not attempt to free arrays if no array were declared in the function
- [x] Fix `mutate` logic
- [ ] Remove `cgutils.printf` calls or wrap then inside `if DEBUG_ENABLE` conditionals

### DIFF
--- a/rbc/irtools.py
+++ b/rbc/irtools.py
@@ -254,6 +254,7 @@ def compile_to_LLVM(functions_and_signatures, target: TargetInfo, debug=False):
                                           flags=flags,
                                           library=main_library,
                                           locals={},
+                                        #   pipeline_class=Pipeline,)
                                           pipeline_class=OmnisciCompilerPipeline)
             make_wrapper(fname, args, return_type, cres)
 
--- a/rbc/irtools.py
+++ b/rbc/irtools.py
@@ -7,7 +7,6 @@ import llvmlite.binding as llvm
 from .targetinfo import TargetInfo
 from .npy_mathimpl import *  # noqa: F403, F401
 from .utils import get_version
-from .omnisci_pipeline import OmnisciCompilerPipeline
 if get_version('numba') >= (0, 49):
     from numba.core import codegen, cpu, compiler_lock, \
         registry, typing, compiler, sigutils, cgutils, \
@@ -239,8 +238,6 @@ def compile_to_LLVM(functions_and_signatures, target: TargetInfo,
     flags = compiler.Flags()
     flags.set('no_compile')
     flags.set('no_cpython_wrapper')
-
-    pipeline_class = OmnisciCompilerPipeline
 
     function_names = []
     for func, signatures in functions_and_signatures:
--- a/rbc/irtools.py
+++ b/rbc/irtools.py
@@ -239,6 +239,8 @@ def compile_to_LLVM(functions_and_signatures, target: TargetInfo, debug=False):
     flags.set('no_compile')
     flags.set('no_cpython_wrapper')
 
+    pipeline_class = OmnisciCompilerPipeline
+
     function_names = []
     for func, signatures in functions_and_signatures:
         for sig in signatures:
@@ -254,8 +256,7 @@ def compile_to_LLVM(functions_and_signatures, target: TargetInfo, debug=False):
                                           flags=flags,
                                           library=main_library,
                                           locals={},
-                                        #   pipeline_class=Pipeline,)
-                                          pipeline_class=OmnisciCompilerPipeline)
+                                          pipeline_class=pipeline_class)
             make_wrapper(fname, args, return_type, cres)
 
     seen = set()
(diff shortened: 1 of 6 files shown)


guilhermeleobas:
Travis is failing but I'm not sure why. It seems like a bug on their side to me.

rbc/irtools.py
```diff
@@ -238,6 +239,8 @@ def compile_to_LLVM(functions_and_signatures, target: TargetInfo, debug=False):
     flags.set('no_compile')
     flags.set('no_cpython_wrapper')
 
+    pipeline_class = OmnisciCompilerPipeline
```
pearu, line 242: We should pass `OmnisciCompilerPipeline` from omniscidb.py via `compile_to_LLVM` its `pipeline_class` argument that needs to be introduced.

rbc/omnisci_array.py
```diff
@@ -90,25 +91,8 @@ class ArrayPointerModel(datamodel.models.PointerModel):
 
     def as_return(self, builder, value):
```
pearu, line 92: This method is not needed and can be removed.

rbc/omnisci_pipeline.py
```diff
@@ -0,0 +1,107 @@
+from numba.core import ir, extending, cgutils
+from numba.core.compiler import CompilerBase, DefaultPassBuilder
+from numba.core.compiler_machinery import FunctionPass, register_pass
+from numba.core.untyped_passes import IRProcessing
```
pearu, line 4: Let's support numba < 0.49 as well:
```
from .utils import get_version
if get_version('numba') >= (0, 49):
    from numba.core import ir, extending, cgutils
    from numba.core.compiler import CompilerBase, DefaultPassBuilder
    from numba.core.compiler_machinery import FunctionPass, register_pass
    from numba.core.untyped_passes import IRProcessing
else:
    from numba import ir, extending, cgutils
    from numba.compiler import CompilerBase, DefaultPassBuilder
    from numba.compiler_machinery import FunctionPass, register_pass
    from numba.untyped_passes import IRProcessing
```

rbc/omnisci_pipeline.py
```diff
@@ -0,0 +1,107 @@
...
+        if not found:
+            return False  # we do not change the IR
+
+        for blk in func_ir.blocks.values():
+            loc = blk.loc
+            scope = blk.scope
+            for ret in blk.find_insts(ir.Return):
+
+                name = 'free_omnisci_array_fn'
+                value = ir.Global(name, free_omnisci_array, loc)
```
pearu, line 80: nit: To eliminate `name`, use
```
value = ir.Global(free_omnisci_array.__name__, free_omnisci_array, loc)
```

rbc/omnisci_pipeline.py
```diff
@@ -0,0 +1,107 @@
...
+    def run_pass(self, state):
+        func_ir = state.func_ir  # get the FunctionIR object
+        found = False
+
+        for blk in func_ir.blocks.values():
+            for stmt in blk.find_insts(ir.Assign):
+                if isinstance(stmt.value, ir.FreeVar) \
+                   and stmt.value.name == 'Array':
+                    found = True
+
```
pearu, line 70: Use:
```
        for blk in func_ir.blocks.values():
            for stmt in blk.find_insts(ir.Assign):
                if isinstance(stmt.value, ir.FreeVar) \
                   and stmt.value.name == 'Array':
                    found = True
                    break
            else:
                continue
            break
```

rbc/omnisci_pipeline.py
guilhermeleobas: I just did that because flake8 complains otherwise, due to the characters line limit

rbc/omnisci_pipeline.py
pearu: I updated the above so that you can eliminate `found`.

rbc/omnisci_pipeline.py
guilhermeleobas: Btw, it seems that intrinsics do not have a `__name__` property.

pearu: ok, nvm then.

rbc/irtools.py
pearu: Remove this line.

rbc/irtools.py
```diff
@@ -7,6 +7,7 @@
 from .targetinfo import TargetInfo
 from .npy_mathimpl import *  # noqa: F403, F401
 from .utils import get_version
+from .omnisci_pipeline import OmnisciCompilerPipeline
```
pearu, line 10: Remove this import.

rbc/irtools.py
guilhermeleobas: I forgot to remove it

PR #69, opened 2020-05-06
Return Omnisci Array

pearu:


rbc/irtools.py
```diff
@@ -295,3 +355,20 @@ def compile_IR(ir):
...
+    """printf that can be called from Numba jit-decorated functions.
+    """
+    if isinstance(format_type, nb_types.StringLiteral):
+        sig = nb_types.void(format_type, arg0, arg1, arg2, arg3, arg4, arg5,
+                            arg6, arg7, arg8, arg9)
+
+        def codegen(context, builder, signature, args):
+            cgutils.printf(builder, format_type.literal_value, *args[1:])
+
+        return sig, codegen
```
guilhermeleobas, line 374: You can create a variadic intrinsic in Numba using `types.BaseTuple`
```suggestion
@extending.intrinsic
def printf(typingctx, format_type, *args):
    if isinstance(format_type, types.StringLiteral):
        sig = types.void(format_type, types.BaseTuple.from_types(args))

        def codegen(context, builder, signature, args):
            cgutils.printf(builder, format_type.literal_value, *args[1:]) 

        return sig, codegen
```

rbc/omnisci_array.py
```diff
@@ -76,7 +89,11 @@ def typer(size, dtype):
 
 @datamodel.register_default(ArrayPointer)
 class ArrayPointerModel(datamodel.models.PointerModel):
-    pass
+
+    def as_return(self, builder, value):
+        # TODO: insert register_buffer_with_executor_rsm call here
+        # print('as_return', value)
```
guilhermeleobas, line 95: Remove this call to `print`

rbc/tests/test_omnisci_array.py
```diff
@@ -282,6 +282,30 @@ def array_setitem_sum(b, c):
...
+def test_array_constructor_return(omnisci):
+    omnisci.reset()
+
+    from rbc.omnisci_array import Array
+    from numba import types
+    from rbc.irtools import printf
+
+    @omnisci('float64[](int32)')
+    def array_return(size):
+        printf("entering array_return(%i)\n", size)
```
guilhermeleobas, line 294: The test passed but I couldn't see the `printf` messages, either locally or on the server. Imagine if there is something to do with stdout not being flushed.

rbc/tests/test_omnisci_array.py
```diff
@@ -282,6 +282,30 @@ def array_setitem_sum(b, c):
...
+        a = Array(size, types.float64)
+        for i in range(size):
+            a[i] = float(i)
+        printf("returning array with length %i\n", len(a))
```
guilhermeleobas, line 298: Same here

rbc/irtools.py
guilhermeleobas: We could send this to numba

pearu: Thanks for the variadic hint.
Sent:
https://github.com/numba/numba/issues/5679

rbc/tests/test_omnisci_array.py
pearu: I can see it when using omniscdb 5.3 and thrift-cpp 0.13.
It is possible that in your case the messages get lost in thift-cpp < 0.13 false-positive error messages.

rbc/omnisci_array.py
pearu: Fixed.

pearu:
The state of this PR is as follows.
When a UDF allocates memory via Array, the memory will be freed only when the UDF returns one of the arrays, except the returned array. The memory of the returned array should be managed by omniscdb server (not implemented yet).

Issue #67, opened 2020-04-29
Replace register_runtime_udf by register_runtime_extension_functions


guilhermeleobas:
On [notebooks/omnisci-demo.ipynb](https://github.com/xnd-project/rbc/blob/master/notebooks/omnisci-demo.ipynb), one uses `register_runtime_udf` but this function is deprecated:

```
 Using register_runtime_udf is deprecated, use register_runtime_extension_functions instead.
```

pearu:
It is not easy to `register_runtime_extension_functions` explicitly because of the signature is not passed as a string (as is in the `register_runtime_udf` case) but as a dictionary of codes that contains the signature type information.

We can move the advance section from the notebook to keep it simple.

guilhermeleobas:
`register_runtime_udf` was removed from the [notebook](https://github.com/xnd-project/rbc/blob/master/notebooks/rbc-omnisci-udf.ipynb)

PR #73, opened 2020-05-14
Update omnisci-demo.ipynb

krunalkharat:
Replaced register_runtime_udf by register_runtime_extension_functions in the existing code😊#67

### DIFF
--- a/notebooks/omnisci-demo.ipynb
+++ b/notebooks/omnisci-demo.ipynb
@@ -319,7 +319,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "omni.thrift_call('register_runtime_udf',\n",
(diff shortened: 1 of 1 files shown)


guilhermeleobas:
Hi @krunalkharat, thanks for the pull request. As [pointed out](https://github.com/xnd-project/rbc/issues/67#issuecomment-621429893) by @pearu, it is not straightforward to use `register_runtime_extension_functions`:

> It is not easy to use register_runtime_extension_functions explicitly because of the signature of a UDF is not passed as a string (as is in the register_runtime_udf case) but as a dictionary of codes that contains the signature type information.
>
> We can move the advance section from the notebook to keep it simple.

Maybe it is just easier to remove this part from the notebook. What do you think, Pearu?

PR #78, opened 2020-05-19
Add fflush function

guilhermeleobas:
This PR adds the `fflush` method. Note the signature of `fflush` is machine/compiler dependent and the one I'm using here matches the one used by qgpu2.


rbc/irtools.py
```diff
@@ -362,6 +362,62 @@ def compile_IR(ir):
...
+        i64, i32, v20,
+    ])
+
+    struct_IO_marker = ir.LiteralStructType([
+        ID_struct_IO_marker.as_pointer(),
+        ID_struct_IO_FILE.as_pointer(),
+        i32,
+    ])
+
+    fflush_fnty = ir.FunctionType(int32_t, [struct_IO_FILE.as_pointer()])
```
pearu, line 401: Could we avoid describing FILE structure and define flush with signature `int (void*)` or similar?
After all, it looks like the only use case for debugging is to call `fflush(NULL)`.

guilhermeleobas: Yeah, that works.

rbc/irtools.py
```diff
@@ -362,6 +362,27 @@ def compile_IR(ir):
...
+
+
+@extending.intrinsic
+def fflush(typingctx):
+    """fflush that can be called from Numba jit-decorated functions.
+    """
+    sig = nb_types.void(nb_types.void)
+
+    def codegen(context, builder, signature, args):
+        _fflush(builder)
```
pearu, line 381: Could you inline `_fflush` as it is not used elsewhere?

guilhermeleobas: It is not possible to call `flush` when generating code. Inlining `_fflush` makes it available to only jitted code. But having it, allows one to call it from a `codegen()` function.

pearu: Sorry for the confusion. By inlining here I mean to insert the `_fflush` Python code to `codegen` so that the function `_fflush` will be eliminated.

guilhermeleobas, line 381:
```suggestion
@extending.intrinsic
def fflush(typingctx):
    """fflush that can be called from Numba jit-decorated functions.
    """
    sig = nb_types.void(nb_types.void)

    def codegen(context, builder, signature, args):
        int8_t = ir.IntType(8)
        fflush_fnty = ir.FunctionType(int32_t, [int8_t.as_pointer()])
        fflush_fn = builder.module.get_or_insert_function(
            fflush_fnty, name='fflush')

        builder.call(fflush_fn, [int8_t.as_pointer()(None)])

    return sig, codegen

```

guilhermeleobas: Like this? And delete `def _fflush()...`?

pearu: exactly.

pearu: Before, `_fflush` was compiler/platform-dependent and verbose, but know it is compiler/platform agnostic and short. Since `_fflush` is used only in codegen, it can be eliminated, imho.

guilhermeleobas: If one removes `_fflush`, we will not be able to call it inside a `codegen()`, only in the body of jitted functions. For instance:

```python
@omnisci('int32(int32)')
def foo(a):
    printf("arg: %d\n", a) # works
    flush() # works
    return a
```

```python
@extending.intrinsic
def foo(typingctx, a):
    sig = types.int32(types.int32)
    def codegen(context, builder, sig, args):
        [arg] = args
        printf("arg: %d\n", arg) # error
        fflush() # error
        # cgutils.printf(builder, "arg: %d\n", arg)
        # _fflush(builder)

        return impl_ret_untracked(context, builder, sig.return_type, arg)

    return sig, codegen
```

pearu: Good point. Then we need two functions.

Suggestion: rename `_fflush` to `codegen_fflush` or `cg_fflush` as it is not a private function.

In the long term, perhaps `cg_fflush` should be exposed via numba as `cgutils.fflush`.

guilhermeleobas: I will add this to Numba as well

pearu: See also the comments in https://github.com/numba/numba/issues/5679 when you are at it.

Issue #76, opened 2020-05-15
Always initialize the array with a default value


guilhermeleobas:
The following code will crash the server due to memory not being initialized:
```python
@omnisci('int64[](int64)')
def foo(sz):
    return Array(sz, 'int64')
```

My proposal would be to initialize arrays with a default value for the type or fill it zeros: `memset(0, sz, len(T) * sz)`

pearu:
I think it should be optional. For instance, if a subsequent code is going to fill array anyway, then calling memset is a waste. We could support the following constructions:
```
Array(sz, dtype)
Array(sz, dtype, fillvalue)
```

On the other hand, could we reuse more of numpy API so that the following constructions would work:
```
np.empty(sz)
np.zeros(sz)
np.ones(sz)
```
etc.? It would be ideal if end-users would not need to use `Array` at all, if possible, so that the defined functions could be backend agnostic and one could call these on numpy arrays as well locally.

guilhermeleobas:
Great idea. I tried to implement `np.zeros` but faced a problem to make numba picking the right implementation. The overload mechanism let us select the correct implementation based on the input types. Since numba already overloads `np.zeros`, any call to it will use the default numba implementation, not the one I did for Omnisci Arrays.

As a workaround, we might have two options:

1. Dig into numba internals and replace the implementation

2. Mimic the NumPy api

```python
def zeros(shape, dtype=types.int64):
    pass

@overload(zeros)
def impl_zeros(shape, dtype=types.int64)
    def impl(shape, dtype=types.int64):
        ...
    return impl


# usage
import omnisci_array as omni  # or something similar
@omnisci('int64[](int64)')
def test_np_zeros(size):
    return omni.zeros(size)
```

pearu:
Yes, the second approach would be preferred. However, consider the example:
```
import numpy as np
@omnisci('int64[](int64)')
def test_np_zeros(size):
    return np.zeros(size)
```
where the `omnisci` decorator re-maps `np` to `rbc.omnisci_numpy_api` (or similar) that implements NumPy API for omnisci arrays. Here is a prototype of how to implement the re-mapping:
```
>>> import numpy as np
>>> def foo():
...     return np.zeros(3)
... 
>>> foo()
array([0., 0., 0.])
>>> class MyNamespace:
...     def zeros(self, size):
...         return [0] * size
... 
>>> omnisci_numpy_api = MyNamespace()
>>> foo.__globals__['np'] = omnisci_numpy_api
[...]

PR #85, opened 2020-05-28
Add np.zeros, np.ones and np.full impl

guilhermeleobas:
Fix https://github.com/xnd-project/rbc/issues/76

Ideally rbc would replace the implementation of `np.zeros` by the one we implemented, but I'm not sure if this is possible inside Numba. I will mark this as a WIP for now.

- [x] `zeros`
- [x] `zeros_like`
- [x] `ones`
- [x] `ones_like`
- [x] `full`
- [x] `full_like`
- [x] `empty`


guilhermeleobas:
appveyor is trying to install numba 0.50.0.rc1 which requires llvmlite >= 0.33. I've tried to force it to use the stable version by setting `numba >= 0.49` on `requirements.txt` but that had no effect.

pearu:
> Any reasons why limiting llvmlite version rule was on setup.py?

The only reason was that it fixed appveyor build. Apparently, only temporarily. A better fix is welcome.

guilhermeleobas:
CI is all green. So, I think we're good.

rbc/tests/test_omnisci_array_functions.py
```diff
@@ -0,0 +1,143 @@
...
+    ('full_like_dtype', 'double[](int8[], double)', ('i1', 3.0), np.full(6, 3, dtype='q')),  # noqa: E501
+    ('ones', 'int32[](int64)', (5,), np.ones(5, dtype=np.int32)),
+    ('ones_dtype', 'double[](int64)', (5,), np.ones(5)),
+    ('ones_like', 'int32[](int32[])', ('i4',), np.ones(6, dtype='i')),
+    ('ones_like_dtype', 'double[](int32[])', ('i4',), np.ones(6, dtype='q')),
+    ('zeros', 'int32[](int64)', (5,), np.zeros(5, dtype=np.int32)),
+    ('zeros_like', 'int32[](int32[])', ('i4',), np.zeros(6, dtype='i')),
+    ('zeros_like_dtype', 'double[](int32[])', ('i4',), np.zeros(6, dtype='q')),
+    ('zeros_dtype', 'double[](int64)', (5,), np.zeros(5)),
+    ('cumsum', 'double[](int32)', (5,), np.arange(1, 6)),
```
pearu, line 125: The `cumsum` crashes omniscidb server (master version), gdb shows:
```
Thread 34 "omnisci_server" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffb1ffb700 (LWP 28468)]
0x00007fffe60c52b9 in rbc::omnisci_array::omnisci_np_cumsum::$3clocals$3e::impl$247($28STRUCT__dPLbK$29$2a) ()
(gdb) bt
#0  0x00007fffe60c52b9 in rbc::omnisci_array::omnisci_np_cumsum::$3clocals$3e::impl$247($28STRUCT__dPLbK$29$2a) ()
#1  0x0000000000000000 in ?? ()

```

guilhermeleobas: I can reproduce this error on omniscidb master

guilhermeleobas: The error happens because this test is calling `np.ones` with the constant value 5 rather than using the `sz` argument. When using the literal value 5 as input, the code is generated and later deleted by the LLVM DCE pass:

```llvm
; Function Attrs: noreturn nounwind
define void @np_cumsum____dPLbKPaiA({ double*, i64, i8 }* nocapture readnone %.1, i32 %.2) local_unnamed_addr #0 {
entry.endif:
  tail call void @llvm.trap()
  unreachable
}
```

Using a non-literal value seems to prevent DCE from deleting everything. I will open an issue to investigate this.

pearu: xref: #109

PR #88, opened 2020-05-31
Fix last_compile regression that caused unnecessary compiles in register

pearu:
This PR enables the last_compile feature that avoids unnecessary function compilations when registering UDFs to the omniscidb server. As a result, the time spent in `test_omnisci.py::test_casting` dropped from 12 to 1.6 seconds.

Also, it adds docstrings to remotejit methods.

### DIFF
--- a/rbc/remotejit.py
+++ b/rbc/remotejit.py
@@ -335,7 +335,7 @@ class RemoteJIT(object):
         # A collection of Caller instances. Each represents a function
         # that have many argument type dependent implementations.
         self._callers = []
-        self._last_ir_map = None
+        self._last_compile = None
         self._targets = None
 
         if local:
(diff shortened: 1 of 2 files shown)


rbc/remotejit.py
```diff
@@ -364,10 +364,57 @@ def reset(self):
...
+    def set_last_compile(self, compile_data):
+        """Save compile data.
+
+        The caller is responsible for discarding previous compiler
+        data by calling `discard_last_compile` method.
+
+        Parameters
+        ----------
+        compile_data : object
+          Compile data can be any Python objecty. When None, it is
```
guilhermeleobas, line 389: `objecty` -> `object`?

pearu: Fixed.

PR #116, opened 2020-07-01
UDTF Column argument support

pearu:
This PR implements:
- The new UDTF API support
- Refactor Omnisci Array to Omnisci Buffer, Array, Column, OutputColumn.
- Various improvements to typesystem.

### DIFF
--- a/rbc/omnisci_backend/omnisci_buffer.py
+++ b/rbc/omnisci_backend/omnisci_buffer.py
@@ -217,12 +217,12 @@ def make_buffer_ptr_type(buffer_type, buffer_ptr_cls):
     numba_type_ptr = buffer_ptr_cls(numba_type, numba_eltype)
 
     ptr_type._params['tonumba'] = numba_type_ptr
-    # todo: add * to typename
-    ptr_type._params['typename'] = buffer_type._params['typename']
+    ptr_type._params['typename'] = buffer_type._params['typename'] + '*'
     return ptr_type
 
 
-def buffer_type_converter(target_info, obj, buffer_typename, buffer_ptr_cls,
+def buffer_type_converter(target_info, obj, typesystem_buffer_type,
+                          buffer_typename, buffer_ptr_cls,
                           extra_members=[]):
     """Template function for converting typesystem Type instances to
     Omnisci Buffer Type instance.
@@ -237,6 +237,7 @@ def buffer_type_converter(target_info, obj, buffer_typename, buffer_ptr_cls,
     ----------
     obj : {str, typesystem.Type}
       Custom atomic type corresponding to Omnisci Buffer.
+    typesystem_buffer_type : {OmnisciBufferType subclass}
     buffer_typename : str
       The name of buffer type.
     buffer_ptr_cls : {BufferPointer subclass}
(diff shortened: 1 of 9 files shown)


rbc/typesystem.py
```diff
@@ -456,6 +475,32 @@ def is_complete(self):
...
+                if not m.is_concrete:
+                    return False
+        elif self.is_function:
+            if not self[0].is_concrete:
+                return False
+            for a in self[1]:
+                if not a.is_concrete:
+                    return False
+        elif self.is_void:
+            pass
```
guilhermeleobas, line 499: Shouldn't this return something?

rbc/omnisci_backend/omnisci_buffer.py
```diff
@@ -0,0 +1,296 @@
...
+    return sig, codegen
+
+
+@extending.intrinsic
+def omnisci_buffer_len_(typingctx, data):
+    sig = types.int64(data)
+
+    def codegen(context, builder, signature, args):
+        data, = args
+        assert data.opname == 'load'
```
guilhermeleobas, line 108: Is this assertion required?

rbc/omnisci_backend/omnisci_buffer.py
```diff
@@ -0,0 +1,296 @@
...
+
+
+@extending.intrinsic
+def omnisci_buffer_getitem_(typingctx, data, index):
+    eltype = data.members[0].dtype
+    sig = eltype(data, index)
+
+    def codegen(context, builder, signature, args):
+        data, index = args
+        assert data.opname == 'load'
```
guilhermeleobas, line 147: Same

rbc/omnisci_backend/omnisci_buffer.py
```diff
@@ -0,0 +1,296 @@
...
+@extending.intrinsic
+def omnisci_buffer_setitem_(typingctx, data, index, value):
+    sig = types.none(data, index, value)
+
+    def codegen(context, builder, signature, args):
+        zero = int32_t(0)
+
+        data, index, value = args
+
+        assert data.opname == 'load'
```
guilhermeleobas, line 193: Same

rbc/omnisci_backend/omnisci_buffer.py
guilhermeleobas, line 108: Is this assertion required?

rbc/omnisci_backend/omnisci_buffer.py
```diff
@@ -0,0 +1,296 @@
...
+
+    def pointer(self):
+        ptr_type = self._params.get('pointer')
+        if ptr_type is not None:
+            return ptr_type
+        return typesystem.Type.pointer(self)
+
+
+class BufferType(types.Type):
+
```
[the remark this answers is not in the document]
pearu: Fixed: added doc string.

rbc/typesystem.py
pearu: The last return statement returns True. The same applies for is_struct and is_function if blocks.

rbc/omnisci_backend/omnisci_buffer.py
pearu: I am uncertain. The subsequent `buf = data.operands[0]` relies on that `data.opname == 'load'`. If the opname is not `load`, using `operands[0]` will lead to unpredictable behaviour and it seems to me that tracking down this possible(?) scenario will be very difficult. So I think the assert statement can be removed if one can be certain that opname is `load`.

PR #117, opened 2020-07-02
Update .travis.yml to run omniscidb server

guilhermeleobas:
trying to setup omniscidb server on travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,8 @@ before_script:
   - conda activate omnisci-ci
   - mkdir data && omnisci_initdb data -f 
   - omnisci_server --enable-runtime-udf --enable-table-functions &
-  - conda deactivate omnisci-ci
+  - conda deactivate
+  - conda activate rbc-ci
 
 script:
   - flake8 .
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,14 +32,14 @@ install:
   - conda list
   - pip install thriftpy2
   - pip install flake8
-  - if [ "$OMNISCIDB_FROM" = "conda" ]; then conda create --quiet --name omnisci-ci omniscidb ; fi
+  - conda create --quiet --name omnisci-ci omniscidb
 
 before_script:
   - python setup.py develop
-  - if [ "$OMNISCIDB_FROM" = "conda" ]; then conda activate omnisci-ci ; fi
-  - if [ "$OMNISCIDB_FROM" = "conda" ]; then mkdir data && omnisci_initdb data -f ; fi
-  - if [ "$OMNISCIDB_FROM" = "conda" ]; then omnisci_server --enable-runtime-udf --enable-table-functions &> omniscidb-output.txt & ; fi
-  - if [ "$OMNISCIDB_FROM" = "conda" ]; then conda deactivate ; fi
+  - conda activate omnisci-ci
+  - mkdir data && omnisci_initdb data -f
+  - omnisci_server --enable-runtime-udf --enable-table-functions 2>&1 > omniscidb-output.txt &
+  - conda deactivate
   - conda activate rbc-ci
 
 script:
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -18,16 +18,23 @@ install:
   - conda config --set always_yes yes --set changeps1 no
   - conda update --quiet --yes conda
   - conda info -a
+  - conda config --add channels conda-forge
   - conda env create --quiet --name rbc-ci --file .conda/environment.yml
```
pearu, line 22: rbc and omniscidb need likely different environments. So, I suggest to try the following approach:
```
conda env create --quiet --name rbc-ci --file .conda/environment.yml
conda env create --quiet --name omnisci-ci omniscidb
conda activate rbc-ci
...
```
and remove the `conda install omniscidb -c conda-forge` below.

Then:
```
before_script:
  - conda activate omnisci-ci
  - mkdir data && omnisci_initdb data -f
  - omnisci_server --enable-runtime-udf --enable-table-functions &
  - conda deactivate omnisci-ci
```

guilhermeleobas:
PR is ready for review. There is an error raised by calcite which I cannot reproduce on omnisci 5.3 or omnisci master:
```
rbc/tests/test_omnisci.py::test_loadtime_udf 2020-07-03 23:56:08 ERROR CalciteServerHandler:process:177 - Validation: org.apache.calcite.runtime.CalciteContextException: From line 1, column 14 to line 1, column 36: No match found for function signature udf_diff2(<NUMERIC>, <NUMERIC>)
```

Also, should we stop pytest on first failure?

pearu:
> Also, should we stop pytest on first failure?

Yes, especially with omnisci tests when the first test crashes the server.

.travis.yml
```diff
@@ -27,7 +29,15 @@ install:
 
 before_script:
   - python setup.py develop
+  - conda activate omnisci-ci
+  - mkdir data && omnisci_initdb data -f 
+  - omnisci_server --enable-runtime-udf --enable-table-functions &
```
pearu, line 34: Suggestion:
```
omnisci_server --enable-runtime-udf --enable-table-functions 2>&1 > omniscidb-output.txt &
```

.travis.yml
```diff
@@ -27,7 +29,15 @@ install:
...
+  - conda activate rbc-ci
 
 script:
   - flake8 .
-  - pytest -v -r s rbc/
\ No newline at end of file
+  - pytest -v -r s rbc/
+
+after_script:
+  - pkill -f omnisci_server
```
pearu, line 43: Continued suggestion:
```
cat omniscidb-output.txt
```

.travis.yml
guilhermeleobas: Do we actually need to pipe the output to a file? It seems that travis doesn't support piping inside an `if` clause:
`if [[ some_condition ]]; omnisci_server --enable-runtime-udf --enable-table-functions 2>&1 > omniscidb-output.txt &`

guilhermeleobas: see: https://travis-ci.org/github/xnd-project/rbc/builds/705525843

pearu: The build log shows command
```
if [ "$OMNISCIDB_FROM" = "conda" ]; then omnisci_server --enable-runtime-udf --enable-table-functions &> omniscidb-output.txt & ; fi
```
but it should be
```
if [ "$OMNISCIDB_FROM" = "conda" ]; then omnisci_server --enable-runtime-udf --enable-table-functions 2>&1 omniscidb-output.txt & ; fi
```

pearu: I think you just had a typo: `&>` ought to be `2>&1`

guilhermeleobas: I have tried that as well. See https://travis-ci.org/github/xnd-project/rbc/jobs/705503755

pearu: Create a `start_omniscidb_bg.sh` script that contains all relevant commands needed to setup and start the omniscidb server.

guilhermeleobas:
I've added conda and docker omnisci versions to travis CI.

Issue #118, opened 2020-07-03
Boolean array is not interpreted as int8 array


guilhermeleobas:
Currently, returning a boolean array from an UDF doesn't work because it is represented as boolean and not as int8.

Code to reproduce
```python
@omnisci('bool[](bool[])')
def invert(inp):
    return np.invert(inp)
```

```
E                   numba.core.errors.TypingError: Failed in nopython mode pipeline (step: nopython frontend)
E                   No conversion from (STRUCT__BPLbK)* to (STRUCT__bPLbK)* for '$12return_value.5', defined at None
```

But the following works:
```python
@omnisci('int8[](bool[])')
def invert(inp):
    return np.invert(inp)
```

Issue #108, opened 2020-06-18
Missing ArrayBool case on ext_arg_type_to_llvm_type


guilhermeleobas:
Currently, it is not possible to return a boolean array from UDF functions.

https://github.com/omnisci/omniscidb/blob/a84b42d324af2f4406abaaf83c465ce11c6039ab/QueryEngine/ExtensionsIR.cpp#L75-L108

pearu:
https://github.com/omnisci/omniscidb-internal/pull/4689 fixes.

PR #119, opened 2020-07-03
Fix returning boolean array

guilhermeleobas:
issue #118

### DIFF
--- a/rbc/omniscidb.py
+++ b/rbc/omniscidb.py
@@ -287,7 +287,7 @@ class RemoteOmnisci(RemoteJIT):
             'bool': typemap['TExtArgumentType']['Bool'],
             'bool*': typemap['TExtArgumentType'].get('PBool'),
             'Array<bool>': typemap['TExtArgumentType'].get(
-                'ArrayBool', typemap['TExtArgumentType']['ArrayInt8']),
+                'ArrayInt8', typemap['TExtArgumentType']['ArrayInt8']),
             'Array<int8_t>': typemap['TExtArgumentType']['ArrayInt8'],
             'Array<int16_t>': typemap['TExtArgumentType']['ArrayInt16'],
             'Array<int32_t>': typemap['TExtArgumentType']['ArrayInt32'],
@@ -458,7 +458,7 @@ class RemoteOmnisci(RemoteJIT):
                         sizer_type = getattr(
                             thrift.TOutputBufferSizeType, sizer)
                         udtfs.append(thrift.TUserDefinedTableFunction(
-                            name + sig.mangling,
+                            name + sig.mangling(bool_is_int8=True),
                             sizer_type, sizer_index,
                             inputArgTypes, outputArgTypes, sqlArgTypes))
                     else:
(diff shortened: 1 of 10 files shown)


pearu:
Could you rebase against the master to fix the Travis CI failure.

guilhermeleobas:
> Could you rebase against the master to fix the Travis CI failure.

Done! Test that I wrote to tackle this is failing and I need to investigate why.

rbc/omniscidb.py
```diff
@@ -287,7 +287,7 @@ def _get_ext_arguments_map(self):
             'bool': typemap['TExtArgumentType']['Bool'],
             'bool*': typemap['TExtArgumentType'].get('PBool'),
             'Array<bool>': typemap['TExtArgumentType'].get(
-                'ArrayBool', typemap['TExtArgumentType']['ArrayInt8']),
+                'ArrayInt8', typemap['TExtArgumentType']['ArrayInt8']),
```
guilhermeleobas, line 290: I've replaced `ArrayBool` by `ArrayInt8` this due to #108

guilhermeleobas:
Tests pass on omniscidb 5.4! Both cuda and cpu versions.

rbc/typesystem.py
```diff
@@ -1118,7 +1121,10 @@ class Boolean1Model(datamodel.models.BooleanModel):
         def get_data_type(self):
             return self._bit_type
 
-    class Boolean8(nb.types.Boolean):
+    class Boolean8(nb.types.Integer):
+
+        def __init__(self, name):
+            super().__init__(name, bitwidth=8, signed=True)
```
guilhermeleobas, line 1127: By inheriting from Integer rather than Boolean, one can get all `int8` overloads for free. Which means, no more errors like the one below:
https://github.com/xnd-project/rbc/blob/6fa215ddbaee50f1a16dd47ed0481ed411631a70/rbc/tests/test_omnisci_math.py#L180-L183

pearu:
Rebased with master and tested with omniscidb 5.3.1 locally. LGTM.

rbc/typesystem.py
```diff
@@ -915,25 +914,29 @@ def _normalize(self, target_info):
...
         if self.is_function:
-            r = self[0].mangle()
-            a = ''.join([a.mangle() for a in self[1]])
+            r = self[0].mangle(bool_is_int8=bool_is_int8)
+            a = ''.join([a.mangle(bool_is_int8=bool_is_int8) for a in self[1]])
             return '_' + r + 'a' + a + 'A'
         if self.is_atomic:
             n = _mangling_map.get(self[0])
             if n is not None:
+                if n == 'b' and bool_is_int8:
```
pearu, line 938: This is problematic as it breaks demangling.

guilhermeleobas: Does #143 fixes this?

pearu: yes

guilhermeleobas: So, we shouldn't map `bool` to `int8` anymore? I'm a bit confuse

PR #130, opened 2020-07-14
fix for pytest 6rc1

guilhermeleobas:
As title

### DIFF
--- a/rbc/tests/test_omnisci_array_math.py
+++ b/rbc/tests/test_omnisci_array_math.py
@@ -129,7 +129,7 @@ def test_omnisci_array_binary_math(omnisci, method, signature, column):
 
     expected = getattr(np, method)(row, row)
 
-    if method == 'power':
+    if method in ['power', 'logaddexp2']:
         assert np.isclose(expected, out, equal_nan=True).all(), 'omni_' + method  # noqa: E501
     else:
         assert np.array_equal(expected, out), 'omni_' + method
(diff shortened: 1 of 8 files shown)


rbc/tests/test_omnisci_array_math.py
```diff
@@ -276,12 +276,7 @@ def test_omnisci_array_unary_math_fns(omnisci, method, signature, column):
 
     row, out = list(result)[0]
     expected = getattr(np, method)(row)
-    _isclose = ['log', 'log2', 'log10', 'log1p', 'sqrt',
-                'arcsin', 'arccos', 'arccosh', 'arctanh']
-    if method in _isclose:
-        assert np.isclose(expected, out, equal_nan=True).all(), 'omni_' + method  # noqa: E501
-    else:
-        assert np.array_equal(expected, out), 'omni_' + method
+    assert np.isclose(expected, out, equal_nan=True).all(), 'omni_' + method  # noqa: E501
```
guilhermeleobas, line 279: Some tests were failing due to precision when using omniscidb 5.3.

guilhermeleobas:
Travis CI is failing because the docker container crashes. I've run the same code on my local docker version and I cannot reproduce the error.

rbc/tests/test_omnisci_array_math.py
pearu: PR https://github.com/xnd-project/rbc/pull/133 has a workaround: skip this test when using 5.3.1+.

guilhermeleobas: Ok, but the test works, it is just a matter of precision (i.e. 1.234567 vs 1.23456789012346)

guilhermeleobas:
Closed in favor of #133

rbc/tests/test_omnisci_array_math.py
pearu: With 5.3.1+, the test failed because `out` contained `None`. With 5.3.0, the test passes here.
Is there still an issue?

guilhermeleobas: I believe `out=None` was in a different place. If I'm not wrong, `out=None` in `max_empty` test.

Issue #134, opened 2020-07-16
test_array_operators/test_array_methods crashes on omniscidb server v5.4 when CUDA is enabled


guilhermeleobas:
In cases where no `sql_execute` were done prior to this test, `omnisci.has_cuda` will be blanked and the test will not be skipped:

#### test_omnisci_array_operators.py
```python
@pytest.mark.parametrize("suffix, signature, args, expected", operator_methods,
                         ids=[item[0] for item in operator_methods])
def test_array_operators(omnisci, suffix, signature, args, expected):

    if omnisci.has_cuda and suffix in ['countOf', 'in', 'not_in']:
        # https://github.com/xnd-project/rbc/issues/107
        pytest.skip(f'operator_{suffix}: crashes CUDA enabled omniscidb server'
                    ' [rbc issue 107]')

    ...
```

#### test_omnisci_array_methods.py
```python
@pytest.mark.parametrize("method, signature, args, expected", ndarray_methods,
                         ids=[item[0] for item in ndarray_methods])
def test_ndarray_methods(omnisci, method, signature, args, expected):
    omnisci.reset()

    if omnisci.has_cuda:
        pytest.skip(
            f'{method}: crashes CUDA enabled omniscidb server [issue 93]')

    ...
```

PR #135, opened 2020-07-16
Fix omnisci.has_cuda is empty in some cases

guilhermeleobas:
Fix issue #134

### DIFF
--- a/rbc/tests/test_omnisci.py
+++ b/rbc/tests/test_omnisci.py
@@ -89,6 +89,10 @@ def omnisci():
 def test_redefine(omnisci):
     omnisci.reset()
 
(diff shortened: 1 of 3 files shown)


guilhermeleobas:
Yes. Actually, cleaning up is just to trigger `retrieve_targets` and set `omnisci.has_cuda` property.

pearu:
I understand the issue. But test_omnisci_array_methods.py and test_omnisci_array.py should be independent. In this PR,  test_omnisci_array_methods.py includes bits from  test_omnisci_array.py while it shouldn't.

guilhermeleobas:
Ops, I see your point and I agree with you. While writing the code, I just copied the code that cleanup the table from `test_omnisci_array.py`. I will update the PR to cleanup a different table.

PR #142, opened 2020-08-03
Add comments to source code

guilhermeleobas:
Added a comment with a TODO and changed code on typesystem.py for code quality.

### DIFF
--- a/rbc/typesystem.py
+++ b/rbc/typesystem.py
@@ -933,10 +933,11 @@ class Type(tuple):
             a = ''.join([a.mangle(bool_is_int8=bool_is_int8) for a in self[1]])
             return '_' + r + 'a' + a + 'A'
         if self.is_atomic:
+            # issue https://github.com/xnd-project/rbc/issues/118
+            if self[0] == 'bool' and bool_is_int8:
+                return _mangling_map.get('int8')
             n = _mangling_map.get(self[0])
             if n is not None:
-                if n == 'b' and bool_is_int8:
-                    return 'B'
                 return n
             n = self[0]
             return 'V' + str(len(n)) + 'V' + n
--- a/rbc/omniscidb.py
+++ b/rbc/omniscidb.py
@@ -286,6 +286,7 @@ class RemoteOmnisci(RemoteJIT):
             'float64*': typemap['TExtArgumentType']['PDouble'],
             'bool': typemap['TExtArgumentType']['Bool'],
             'bool*': typemap['TExtArgumentType'].get('PBool'),
+            # TODO: Replace ArrayInt8 by ArrayBool after issue #108 is solved
             'Array<bool>': typemap['TExtArgumentType'].get(
                 'ArrayInt8', typemap['TExtArgumentType']['ArrayInt8']),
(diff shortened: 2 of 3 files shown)


rbc/typesystem.py
```diff
@@ -933,10 +933,11 @@ def mangle(self, bool_is_int8=None):
             a = ''.join([a.mangle(bool_is_int8=bool_is_int8) for a in self[1]])
             return '_' + r + 'a' + a + 'A'
         if self.is_atomic:
+            # issue https://github.com/xnd-project/rbc/issues/118
+            if self[0] == 'bool' and bool_is_int8:
```
pearu, line 937: This seems to be unnecessary after #143 .

guilhermeleobas: That is still relevant, no? Without this, `bool` will be mapped to `b` rather than `B` (shortcut for `int8`).
```python
_mangling_map = dict(
    void='v', bool='b',
    char8='c', char16='z', char32='w',
    int8='B', int16='s', int32='i', int64='l', int128='q',
    uint8='U', uint16='S', uint32='I', uint64='L', uint128='Q',
    float16='h', float32='f', float64='d', float128='x',
    complex32='H', complex64='F', complex128='D', complex256='X',
    string='t',
)
```

pearu: No. typesystem has only one boolean type: `b`.

Note that the flag `bool_is_int8` is used only in conversion *to numba* type and is (omnisci llvm) implementation-specific flag. So, conversion *from numba* undoes the effect of `bool_is_int8`: `boolean8` is mapped back to `b`.
In fact, `bool` should never map to `B` because `B` is a shortcut to 8-bit integer and in the typesystem level, boolean and integers are different types. Only at the implementation level, the boolean type is the same as LLVM `i1` or LLVM `i8`.

Numba implements boolean as LLVM `i1` while for omnisci boolean is LLVM `i8`. So in some sense, the flag `bool_is_int8` means that for omnisci, all instances of LLVM `i1` that Numba generates should be replaced with LLVM `i8`.

Stated differently, we have three levels of types: typesystem, numba, LLVM IR. We have the following flows of converting typesystem boolean to implementation specific boolean:
```
typesystem bool -> numba boolean1 -> LLVM i1  when bool_is_int8=False
typesystem bool -> numba boolean8 -> LLVM i8  when bool_is_int8=True
```
Notice that boolean becomes integer only at the LLVM level. At the numba level, boolean remains a Boolean instance.

Finally, since computing the mangled string is arbitrary (as long as the result can be reversed), I think `bool_is_int8` should not be used in the mangling algorithm to avoid unifying `B` and `b`. 
I'll test if eliminating bool_is_int8 from mangling breaks anything..

pearu: The PR https://github.com/xnd-project/rbc/pull/145 eliminates bool_is_int8 from the mangling algorithm.

guilhermeleobas: Makes sense, thanks for the explanation.

guilhermeleobas:
I will close this PR as it doesn't makes sense anymore.

PR #143, opened 2020-08-03
Keep boolean8 and int8 separate in numba side (fixes issue 140)

pearu:


guilhermeleobas:
Tests passed on omniscidb 5.4 cpu and cuda version

PR #144, opened 2020-08-03
enable boolean8 tests and call functions from the "omni" pkg

guilhermeleobas:
- explicit cast of boolean8 to bool
- call functions from omni. pkg rather than np.
- replace int8[] by bool[] in some tests

This PR requires a small change in omniscidb to handle boolean values: https://github.com/omnisci/omniscidb-internal/pull/4689

### DIFF
--- a/rbc/omnisci_backend/numpy_ufuncs.py
+++ b/rbc/omnisci_backend/numpy_ufuncs.py
@@ -9,7 +9,18 @@ else:
 
 
 def determine_dtype(a, dtype):
-    return a.eltype if dtype is None else dtype
+    if isinstance(a, ArrayPointer):
+        return a.eltype if dtype is None else dtype
+    else:
+        return a if dtype is None else dtype
+
+
+def determine_input_type(argty):
+    if argty == typesystem.boolean8:
+        return bool
+    else:
+        return argty.eltype if isinstance(argty, ArrayPointer) else argty
+
 
 
 def overload_elementwise_binary_ufunc(ufunc, name=None, dtype=None):
@@ -21,13 +32,18 @@ def overload_elementwise_binary_ufunc(ufunc, name=None, dtype=None):
     globals()[name] = ufunc
 
     def binary_ufunc_impl(a, b):
+        typA = determine_input_type(a)
+        typB = determine_input_type(b)
+
         # XXX: raise error if len(a) != len(b)
         @extending.register_jitable
         def binary_impl(a, b, nb_dtype):
             sz = len(a)
             x = Array(sz, nb_dtype)
             for i in range(sz):
-                x[i] = nb_dtype(ufunc(a[i], b[i]))
+                cast_a = typA(a[i])
+                cast_b = typB(b[i])
+                x[i] = nb_dtype(ufunc(cast_a, cast_b))
             return x
 
         @extending.register_jitable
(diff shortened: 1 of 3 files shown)


guilhermeleobas:
PR is ready for review. Wait for omnisci/omniscidb-internal/pull/4689 to be merged before merging this one.

rbc/omnisci_backend/numpy_ufuncs.py
```diff
@@ -9,7 +9,17 @@
 
 
 def determine_dtype(a, dtype):
-    return a.eltype if dtype is None else dtype
+    if isinstance(a, ArrayPointer):
+        return a.eltype if dtype is None else dtype
+    else:
+        return a if dtype is None else dtype
```
pearu, line 15: `return a` is suspicious: the function ought to return type instance..

rbc/omnisci_backend/numpy_ufuncs.py
```diff
@@ -9,7 +9,17 @@
...
+
+
+def determine_input_type(argty):
+    if argty == typesystem.boolean8:
+        return bool
+    else:
+        return argty.eltype if isinstance(argty, ArrayPointer) else argty
```
pearu, line 22: What are the allowed return values of `determine_dtype` and `determine_input_type`? Python types, Numba types, anything else? I presume to return typesystem types is not allowed.

rbc/tests/test_omnisci_math.py
```diff
@@ -162,33 +164,43 @@ def omnisci():
...
+        fn.__name__ = fn_name
 
-    if fn_name in ['logical_or', 'logical_xor', 'logical_and', 'logical_not']:
+    if available_version[:2] <= (5, 3) and fn_name in \
+            ['logical_or', 'logical_xor', 'logical_and', 'logical_not']:
         # Invalid use of Function(<ufunc 'logical_or'>) with
         # argument(s) of type(s): (boolean8, boolean8)
-        pytest.skip(f'using boolean8 as {fn_name} argument not implemented')
+        pytest.skip(
+            f'using boolean8 as {fn_name} argument not implemented for'
```
pearu, line 195: suggest rewording: `"using boolean arguments requires omniscidb v 5.4 or newer (got {available_version})"`

rbc/omnisci_backend/numpy_ufuncs.py
guilhermeleobas: `determine_dtype` and `determine_input_type` must return numba types.

* `determine_input_type`: Only exists to explicitly convert `boolean8` values to `bool`. Without this, one would have to write numpy functions to work with `boolean8` args (see #146). 
* `determine_dtype`: Explicit convert the output to a dtype.

> I presume to return typesystem types is not allowed.

Yes, you're right.

rbc/omnisci_backend/numpy_ufuncs.py
guilhermeleobas: `a` ought to be a type. I can add a check at the beginning of this function to make sure it is a type. What do you think?

pearu: Perhaps introduce a decorator on  determine_dtype and determine_input_type that checks that the return type is numba type?

pearu: Can you point the case where `return a` is effective?

guilhermeleobas: Yes. Look at the implementation of `np.negative(<int64>)` for instance. In this case `return a` returns `int64` since the dtype is `None`. But on `np.logical_not`, the output needs to be a boolean value, and we enforce this constraint using the dtype argument.

PR #145, opened 2020-08-04
Eliminate bool_is_int8 from mangling algorithm

pearu:
`bool_is_int8` ought to be used only when converting typesystem types to numba types.

### DIFF
--- a/rbc/irtools.py
+++ b/rbc/irtools.py
@@ -247,7 +247,7 @@ def compile_to_LLVM(functions_and_signatures, target: TargetInfo,
     function_names = []
     for func, signatures in functions_and_signatures:
         for sig in signatures:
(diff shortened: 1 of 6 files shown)


pearu:
@guilhermeleobas, I think we should keep typesystem types implementation-agnostic.

A boolean value has two states (True, False) and hence LLVM `i1` would be the most appropriate implementation type for booleans. This is what also numba advocates.
(*) Although, after compilation LLVM `i1` values likely occupy the same amount of memory as LLVM `i8`.

Omnisci uses LLVM `i8` to implement boolean type because in omnisci a boolean value has three states: True, False, unspecified. This is the main reason why omnisci uses `i8` for boolean types, (*) is a secondary reason.

Issue #152, opened 2020-08-10
Implement support for signbit


pearu:
numba uses numpy implementation of signbit that is included in numpy RT library. However, numpy RT not available for omniscidb server.

Possible options:
- implement signbit support using the signbit functions from math library
- expose numpy RT to omniscidb server, it must not include any Python functions.bit.

The approach taken must be applicable for other functions that have similar implementations to signbit.

guilhermeleobas:
For ufuncs, we ought to replace the implementation on `npy_mathimpl.py` like we did for `np.logaddexp` and `np.logaddexp2`. The reason for this is that once the function is registered by Numba, it becomes impossible (without changing internals) to replace the implementation. So, on `npy_mathimpl.py` we change the implementation that Numba will use. 

https://github.com/xnd-project/rbc/blob/8d576bddb5778f40f521ab1362b07112edae24f2/rbc/npy_mathimpl.py#L73-L85

pearu:
> For ufuncs, we ought to replace the implementation on `npy_mathimpl.py` ...

Does it break numba?

guilhermeleobas:
> > For ufuncs, we ought to replace the implementation on `npy_mathimpl.py` ...
> 
> Does it break numba?

No, but the NumPy implementation may be more precise than ours. For instance, consider the implementation for `np.nextafter`:

https://github.com/numpy/numpy/blob/6d4895cc9fe614cf3164891b1f502453312f64ca/numpy/core/src/npymath/ieee754.c.src#L400-L449

guilhermeleobas:
Closed by https://github.com/xnd-project/rbc/pull/154

PR #154, opened 2020-08-10
Add np.signbit, np.ldexp

guilhermeleobas:
xref: #152

### DIFF
--- a/rbc/npy_mathimpl.py
+++ b/rbc/npy_mathimpl.py
@@ -83,7 +83,9 @@ def np_signbit_impl(context, builder, sig, args):
 def np_ldexp_impl(context, builder, sig, args):
 
     def impl(x1, x2):
-        return x1 * 2 ** x2
+        if x2 < 0:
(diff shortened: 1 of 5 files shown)


rbc/npy_mathimpl.py
```diff
@@ -70,6 +70,26 @@ def impl(x, y):
     return context.compile_internal(builder, impl, sig, args)
 
 
+def np_signbit_impl(context, builder, sig, args):
+
+    def impl(x):
+        if x < 0:
+            return True
+        return False
```
pearu, line 78: Float zero has also signbit, e.g. -0.0 and 0.0 are different float values. See how signbit is implemented in numpy: https://github.com/numpy/numpy/blob/master/numpy/core/src/npymath/_signbit.c . 
Is it possible to implement the union trick here?

guilhermeleobas: I copied this code on godbold.com and implemented the IR it outputs. See: https://godbolt.org/z/P5Wj4b

Issue #160, opened 2020-08-20
omniscidb assigns the same column value to every column argument


guilhermeleobas:
Example:

```python
@omnisci('int32(Column<double>, Column<double>, '
        ' int64|sizer=RowMultiplier, OutputColumn<double>)')
def sample_udtf_fn(S, X, m, out):
    input_row_count = len(X)
    for i in range(input_row_count):
        out[i] = S[i] + X[i]
    return m * input_row_count

omnisci.sql_execute(
    'CREATE TABLE IF NOT EXISTS test_table (S DOUBLE, X DOUBLE);')

omnisci.sql_execute('INSERT INTO test_table VALUES (10, 20)')
omnisci.sql_execute('INSERT INTO test_table VALUES (15, 25)')
omnisci.sql_execute('INSERT INTO test_table VALUES (17, 27)')

_, result = omnisci.sql_execute(
    'select * from table(sample_udtf_fn('
    ' cursor(select S from test_table),'
    ' cursor(select X from test_table),'
    ' 1));'
    .format(**locals()))

print(list(result))
# [(20.0,), (30.0,), (34.0,)]
```

The expected output is `[(30.0,), (40.0,), (44.0,)]`

guilhermeleobas:
Closed by https://github.com/omnisci/omniscidb-internal/pull/4747.

PR #161, opened 2020-08-20
Add black-scholes test

guilhermeleobas:
As title. UDTF example do not work as reported on #160.


guilhermeleobas:
UDTF tests works with the fix from https://github.com/omnisci/omniscidb-internal/pull/4747

pearu (approved):
LGTM

table_load_columnar provides an efficient way to load data to omniscidb server. So, after resolving https://github.com/xnd-project/rbc/issues/165 we can enable the test to the full data in the used cvs file.

PR #163, opened 2020-08-23
Enable array support in load_table_columns

pearu:
As in title. In addition, the PR tests multi column input UDTF that requires https://github.com/omnisci/omniscidb-internal/pull/4747


pearu:
No, we can merge immediately as the CI uses omniscidb 5.3 while the corresponding test that requires omniscidb-internal 4747 is enabled only for omniscidb 5.4. load_table_columns should work with all omniscidb versions.

PR #166, opened 2020-08-25
Add black-scholes notebook

guilhermeleobas:
As title.


pearu (approved):
LGTM.
I adjusted notebook text a little and updated to use the latest rbc features: RowMultiplier type.

PR #183, opened 2020-09-11
Fix notebooks

guilhermeleobas:
This PR resolves the following items:
- Ensure notebooks are executable
- Update notebooks with the latest features of RBC (i.e. `load_columnar_data`)
- Remove `remote_numba.ipynb` notebook since it uses old code and its use case is covered by other notebooks.


pearu (approved):
LGTM

PR #195, opened 2020-09-20
Support multiple columns in a single cursor argument of a UDTF.

### LINKED ISSUE #173: Support single cursor for multiple input columns
The following example should work:
```

    @omnisci('int32(Column<double>, Column<bool>, RowMultiplier, OutputColumn<double>)')
    def mask_zero(x, b, m, y):
        for i in range(len(x)):
            if b[i]:
                y[i] = 0.0
            else:
                y[i] = x[i]
        return len(x)
```
Currently, one can use:
```
omnisci.sql_execute(f'select * from table(mask_zero(cursor(select f8 from {omnisci.table_name}), cursor(select b from {omnisci.table_name}), 1))')
```
which is inefficient, unreliable, and not user-friendly.

One should be able to use:
```
omnisci.sql_execute(f'select * from table(mask_zero(cursor(select f8, b from {omnisci.table_name}), 1))')
```
that fails with:
```
No match found for function signature mask_zero(<CURSOR>, <NUMERIC>)
```

pearu:
Fixes #173 .
Requires omniscidb 5.5 and https://github.com/omnisci/omniscidb-internal/pull/4856 .

Example:
```
@omnisci('int32(Cursor<double, bool>, RowMultiplier, OutputColumn<double>)')
    def mask_zero(x, b, m, y):
        for i in range(len(x)):
            y[i] = x[i] if b[i] else 0.0
        return len(x)

omnisci.sql_execute('select * from table(mask_zero(cursor(select x, b from mytable), 1)')
```

### DIFF
--- a/rbc/omnisci_backend/omnisci_column.py
+++ b/rbc/omnisci_backend/omnisci_column.py
@@ -110,7 +110,17 @@ def cursor_type_converter(target_info, obj):
         raise NotImplementedError(type(obj))
     name, params = obj.get_name_and_parameters()
     if name == 'Cursor':
-        cursor_type = OmnisciCursorType(*tuple(typesystem.Type.fromstring(
-            p, target_info=target_info) for p in params))
+        new_params = []
+        for p in params:
+            p = typesystem.Type.fromstring(p, target_info=target_info)
+            if p.is_float or p.is_int or p.is_bool:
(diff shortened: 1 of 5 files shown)


guilhermeleobas (approved):
LGTM

Issue #175, opened 2020-09-10
Support redefining UDTFs


pearu:
One can redefine UDFs but not UDTFs.

UDTFs are registered differently from UDFs, the entry to relevant code is:
https://github.com/omnisci/omniscidb-internal/blob/master/ThriftHandler/DBHandler.cpp#L6157

pearu:
The redefinition of UDTF works when changing the body of the UDTF.

When the UDTF argument types are changed, say, from `double` to `int32`, the output of a UDTF contains garbage.

pearu:
This issue can be resolved once https://github.com/omnisci/omniscidb-internal/pull/4906 and https://github.com/xnd-project/rbc/pull/199 are merged.

PR #199, opened 2020-10-07
Tests for table function overload/redefine support

pearu:
Resolves issue #175 .

Requires https://github.com/omnisci/omniscidb-internal/pull/4906


guilhermeleobas (approved):
LGTM

PR #213, opened 2020-11-12
Test for lazy linking of UDF/UDTFs

### LINKED ISSUE #186: Implement lazy linking of UDF/UDTFs
Lazily link udfs/udtfs so that we don’t force all kernels to CPU if one UDTF or UDF cannot run on GPU

guilhermeleobas:
Fix #186


guilhermeleobas:
Closing this since the issue is related to C++ UDFs

PR #219, opened 2020-11-19
Make TargetInfo globally available

### LINKED ISSUE #97: Make target_info object globally available
It would be good if we have the `target_info` object globally available. In the current approach, we only have access to it through the typing context. For instance, when implementing `np.ones`, `nb_dtype` must be a numba dtype and if `dtype` is a type string, `nb_dtype` will hold the wrong value.

```python
@extending.overload(ones)
def omnisci_np_ones(shape, dtype=None):

    if dtype is None:
        nb_dtype = nb_types.double
    else:
        nb_dtype = dtype
        # it would be good if we could use the typesystem here:
        # nb_dtype = typesystem.Type(target_info, dtype).tonumba()

    def impl(shape, dtype=None):
        return full(shape, 1, nb_dtype)
    return impl
```

guilhermeleobas:
Fixes #97

### DIFF
--- a/rbc/omniscidb.py
+++ b/rbc/omniscidb.py
@@ -830,89 +830,85 @@ class RemoteOmnisci(RemoteJIT):
         fid = 0  # UDF/UDTF id
         udfs, udtfs = [], []
         for device, target_info in self.targets.items():
-            # Set a global target_info instance
-            TargetInfo.set_instance(target_info)
+            with target_info:
+                udfs_map, udtfs_map = {}, {}
+                functions_and_signatures = []
+                function_signatures = defaultdict(list)
+                for caller in reversed(self.get_callers()):
+                    signatures = {}
+                    name = caller.func.__name__
+                    if name in self.forbidden_names:
+                        raise ForbiddenNameError(
+                            f'\n\nAttempt to define function with name `{name}`.\n'
+                            f'As a workaround, add a prefix to the function name '
+                            f'or define it with another name:\n\n'
+                            f'   def prefix_{name}(x):\n'
+                            f'       return np.trunc(x)\n\n'
+                            f'For more information, see: '
+                            f'https://github.com/xnd-project/rbc/issues/32')
+                    for sig in caller.get_signatures(target_info):
+                        i = len(function_signatures[name])
+                        if sig in function_signatures[name]:
+                            if self.debug:
+                                f2 = os.path.basename(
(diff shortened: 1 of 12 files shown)


rbc/targetinfo.py
```diff
@@ -8,6 +8,19 @@ class TargetInfo(object):
     """Holds target device information.
     """
 
+    @classmethod
+    def set_instance(cls, instance):
+        cls._instance = instance
+
+    @classmethod
+    def instance(cls):
+        assert cls._instance is not None, "No TargetInfo instance was set prior to this point."
```
pearu, line 17: Use:
```
if cls._instance is None:
    raise RuntimeError('Target not specified.')
```

rbc/omniscidb.py
```diff
@@ -830,6 +830,9 @@ def is_udtf(sig):
         fid = 0  # UDF/UDTF id
         udfs, udtfs = [], []
         for device, target_info in self.targets.items():
+            # Set a global target_info instance
+            TargetInfo.set_instance(target_info)
+
```
pearu, line 835: This construction begs for using the `with` statement. Let's make `TargetInfo` a context manager (define methods `__enter__`, `__exit__`) so that one can use:
```
with target_info:
    ...
```
instead of
```
TargetInfo.set_instance(target_info)
...
TargetInfo.clear_instance()
```

pearu (changes requested):
Great work!

Notice that with this PR, a target info instance is globally available. This means that (almost) all `target_info` arguments in various methods/functions are unnecessary: instead of passing the `TargetInfo` instance through many levels of function calls that even don't need `target_info`, one can use `TargetInfo.instance()` to access the instance where it is actually used.

Could you revisit the `typesystem.Type.from...` methods and make `target_info` argument optional. For instance:
```
def fromobject(self, obj, target_info=None):
    if target_info is None:
        target_info = self.instance()
```
And then eliminate `target_info` arguments from functions/methods that just pass on the `TargetInfo` to other function/method calls.
Also, make sure that if a `TargetInfo` instance is stored in some data structure then it is really needed.

pearu (approved):
@guilhermeleobas , I have removed more unnecessary `target_info` throughputs and enabled using `TargetInfo()` to retrieve the global target info instance. Please review as well.

guilhermeleobas:
LGTM

PR #223, opened 2020-11-24
Add support for table_id attribute [ON HOLD]

### LINKED ISSUE #177: Attach table info to Column inputs of UDTFs
Something like the following should work
```
@omnisci('int32(Column<double>, Column<double>, RowMultiplier, OutputColumn<double>)')
def foo(x, y, m, z):
    if x.table_id == y.table_id:
        # do something
    else:
        # do something different
    <return z size> 
```

Risks:
- the above makes sense for queries like
  ```
  select * from table(foo(mytable.x, mytable.y, 1))
  ```
  but does it for queries like
  ```
  select * from table(foo(cursor(select x, y from mytable), 1))
  select * from table(foo(cursor(select x from mytable), cursor(select x from mytable2), 1))
  ```
  where the table info may get lost due to usage of `cursor` or when the input column is an output of another UDTF?

guilhermeleobas:
Fixes #177 

Requires https://github.com/omnisci/omniscidb-internal/pull/5012

### DIFF
--- a/rbc/tests/test_omnisci_column_basic.py
+++ b/rbc/tests/test_omnisci_column_basic.py
@@ -72,6 +72,61 @@ def test_sizer_row_multiplier_orig(omnisci):
         assert r == (float((i % 5) * 2),)
 
 
+@pytest.mark.skipif(
+    available_version < (5, 5),
+    reason=(
+        "test requires omniscidb with table_id attribute"
+        " support (got %s) [issue 124]" % (
+            available_version,)))
+def test_input_column_table_id(omnisci):
+    omnisci.reset()
+
+    @omnisci('int32(Column<double>, RowMultiplier, OutputColumn<double>)')
+    def udtf_table_id(x, m, z):
+        input_row_count = len(x)
+        for i in range(input_row_count):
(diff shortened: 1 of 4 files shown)


rbc/tests/test_omnisci_column_basic.py
```diff
@@ -72,6 +79,61 @@ def my_row_copier_mul(x, m, y):
...
+        return m * input_row_count
+
+    descr, result = omnisci.sql_execute(
+        'select * from table(udtf_table_id('
+        'cursor(select f8 from {omnisci.table_name}),'
+        '1));'
+        .format(**locals()))
+
+    for i, r in enumerate(result):
+        assert r == (-1.0,)
```
pearu, line 134: Suggest removing `test_input_column_table_id` and `test_output_column_table_id` in favor of `test_column_get_table_id`.

pearu (changes requested):
Suggest using `test_column_get_table_id` as it covers all use cases in a cleaner way.

Ideally, I think, table_id should contain the id of the table where the column is stored.

Issue #192, opened 2020-09-20
Support strings


pearu:
Perhaps already have string dictionary support by inputting key_for_string, but could make nicer ux around this

pearu:
See Slack discussions for ideas:

1. https://quansight.slack.com/archives/C01CAU0U1FH/p1605526178096400

pearu:
Related PRs:
- https://github.com/omnisci/omniscidb-internal/pull/5492
- https://github.com/xnd-project/rbc/pull/322

guilhermeleobas:
Related PR:
* https://github.com/xnd-project/rbc/pull/322

guilhermeleobas:
TextEncodingNone and TextEncodingDict types are supported.

pearu:
Tests in rbc/tests/heavyai/test_text_encoding.py cover the cases where TextEncodingDict variables are passed around but we don't have tests for accessing the content of TextEncodingDict variables. In short, we should be able to define a UDF that translates a TextEncodingDict variable to TextEncodingNone return value, for instance. And vice-versa, checking if TextEncodingNone variable can be converted to TextEncodingDict return value (I imagine if the input string is not in the strings dictionary, the conversion should fail, or add a new item to the dict).

guilhermeleobas:
PR #531 adds support for python strings

PR #225, opened 2020-11-26
Support TEXT ENCODING NONE as Bytes

pearu:
As in title.

In addition:
- Introduce process_callable method that enables detecting user errors when not importing Array and other types
- Continue moving Array features to Buffer
- Various clean ups
- Introduce `devices` kw argument to remotejit decorator so that the desired target devices can be specified manually.

Requires: https://github.com/omnisci/omniscidb-internal/pull/5017

Related to #192

### DIFF
--- a/rbc/omniscidb.py
+++ b/rbc/omniscidb.py
@@ -718,7 +718,6 @@ class RemoteOmnisci(RemoteJIT):
 
         consumed_index = 0
         name = caller.func.__name__
-        type_to_extarg = self.type_to_extarg
         for i, a in enumerate(orig_sig[1]):
             annot = a.annotation()
             _sizer = annot.get('sizer', unspecified)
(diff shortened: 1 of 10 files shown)


rbc/omniscidb.py
```diff
@@ -703,6 +715,7 @@ def _make_udtf(self, caller, orig_sig, sig):
 
         consumed_index = 0
         name = caller.func.__name__
+        type_to_extarg = self.type_to_extarg
```
guilhermeleobas, line 718: Any particular reason for assign this rather than just use `self.type_to_extarg`?

rbc/omniscidb.py
```diff
@@ -907,6 +908,7 @@ def is_udtf(sig):
                     llvm_function_names.append(f.name)
 
                 device_ir_map[device] = str(llvm_module)
+                # print(device_ir_map[device])
```
guilhermeleobas, line 911: remove this comment

rbc/targetinfo.py
```diff
@@ -305,6 +306,27 @@ def has_cpython(self):
...
+                if self.is_gpu:
+                    return True
+            elif d == 'cuda':
+                if self.driver[0] == 'CUDA':
+                    return True
+            else:
+                warnings.warn(
+                    f'Ignoring unknown desired device specification: {d}')
+        return False
+
```
guilhermeleobas, line 329: `supports_devices` needs to check if every device in `desired_devices` is supported? or if at least one device is supported? If the former is the case, then, I guess this function is wrong. For instance, consider the following case: `self.supports_device(['cpu', 'gpu', 'a random string'])`. Even though target may support `cpu` and `gpu`, `a random string` is unknown.

pearu: It's the latter. Notice that the `TargetInfo` instance `self` represents a single target device, cpu or gpu. The `self.supports_devices` method tests if the `self` target device is in the `desired_devices` list.

So, the doc string is wrong and the naming of the method could be better. Here's a plan:
```
def contained_in(self, desired_devices):
    """Checks if target is in a set of desired devices.

    Parameters
    ----------
    desired_devices: list, None
      Devices names that can be a subset of `['cpu', 'gpu', 'cuda']`.

    Returns
    -------
    is_contained : bool
      True, if the target device name is listed in `desired_devices`
      or `desired_devices` is None. False, otherwise.
    """
```
Any suggestions?

guilhermeleobas: `contained_in` seems a good name. Or maybe `supports_any_device`

pearu: Moved this to https://github.com/xnd-project/rbc/pull/228/files

pearu:
Blocked by https://github.com/xnd-project/rbc/pull/228

rbc/omniscidb.py
pearu: No. Now using `self.type_to_extarg` instead.

rbc/omniscidb.py
pearu: Fixed.

PR #228, opened 2020-12-04
Introduce TargetInfo.check_enabled and RemoteJIT.preprocess_callable

pearu:
This feature allows specifying devices for which a UDF/UDTF will be defined. For example:

```
@omnisci('int(int)', devices = ['cpu'])
def foo_executes_only_on_cpu(x):
    return x + 1

@omnisci('int(int)', devices = ['gpu'])
def bar_executes_only_on_gpu(x):
    return x + 1

@omnisci('int(int)', devices = ['cpu', 'gpu'])
def fun_executes_on_cpu_and_gpu(x):
    return x + 1

@omnisci('int(int)')
def fun_executes_on_all_devices(x):
    return x + 1
```
The values `'gpu'` and `'cuda'` are equivalent (at least for now).

### DIFF
--- a/rbc/remotejit.py
+++ b/rbc/remotejit.py
@@ -55,6 +55,11 @@ class Signature(object):
         assert isinstance(remotejit, RemoteJIT), type(remotejit)
         self.remotejit = remotejit
         self.signatures = []
+        self.signature_devices = {}
+
+    @property
+    def debug(self):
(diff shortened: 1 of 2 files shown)


pearu:
Merging this to remove blocker for implementing tests for #186 .

PR #230, opened 2020-12-04
Add is_null method for column access

### LINKED ISSUE #188: Null check on column accessor
Either `is_null(col_var, row_idx)`, or `col_var[row_idx].is_null()`.
Can we leverage existing codegen for this?

guilhermeleobas:
Fix #188

Requires: https://github.com/omnisci/omniscidb-internal/pull/5104 (was https://github.com/omnisci/omniscidb-internal/pull/5074)

Edit: test passes on both cpu and gpu versions of omniscidb

### DIFF
--- a/rbc/irtools.py
+++ b/rbc/irtools.py
@@ -182,7 +182,7 @@ class RemoteGPUTypingContext(typing.Context):
 # Code generation methods
 
 
-def make_wrapper(fname, atypes, rtype, cres, verbose=False):
+def make_wrapper(fname, atypes, rtype, cres, target: TargetInfo, verbose=False):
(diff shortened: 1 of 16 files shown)


pearu (changes requested):
See https://github.com/omnisci/omniscidb-internal/pull/5104#pullrequestreview-553600381 .
Need a reliable way to serialize floating-point values.

Notice that we don't need to materialize the null values from the rbc side when using LLVM IR bitcast-to.

guilhermeleobas:
Hi Pearu, I've change the serialization method to use the hex representation of the float rather than `to_string(float)`.

pearu:
Replacing this PR with #239

PR #231, opened 2020-12-11
Add test for support columns with different sizes as inputs to UDTFs

### LINKED ISSUE #176: Support columns with different sizes as inputs to UDTFs
The following example should work:
```
@omnisci('int32(Column<double>, Column<double>, RowMultiplier, OutputColumn<double>)')
def convolve(x, kernel, m, y):
    for i in range(len(x)):
        for j in range(len(kernel)):
            k = i + j
            if (k < len(x)):
                y[k] += kernel[j] * x[k]
    return len(x)
```
where `len(kernel) < len(x)`.

Currently, all input columns are assumed to have the same length which is fine if the input columns are from the same table but one might want to pass columns from different tables with different row sizes. For instance,
```
omnisci.execute('''
    select * from table(convolve(cursor(select x from datatable),
                                 cursor(select kernel from kerneltable), 1))')
```

guilhermeleobas:
Fixes #176 

Requires https://github.com/omnisci/omniscidb-internal/pull/5086

### DIFF
--- a/rbc/tests/test_omnisci_column_basic.py
+++ b/rbc/tests/test_omnisci_column_basic.py
@@ -501,8 +501,6 @@ def test_redefine(omnisci):
         assert result[i][0] == f8[i] + 2
 
 
-# remove xfail after rbc pull request #232 is merged
-@pytest.mark.xfail(reason="rbc pr #232")
 @pytest.mark.skipif(
     available_version < (5, 5),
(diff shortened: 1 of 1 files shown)


guilhermeleobas:
https://github.com/omnisci/omniscidb-internal/pull/5086 was merged

PR #229, opened 2020-12-04
Implement tests for omniscdb-internal PR 5026

pearu:
Requires https://github.com/omnisci/omniscidb-internal/pull/5026

PR #232, opened 2020-12-14
Change pytest.raises match message depending on omniscidb version

guilhermeleobas:
The error message when there is not a match changed in OmniSciDB 5.6. This small PR updates RBC to work with this change.


pearu:
This issue is already resolved in https://github.com/xnd-project/rbc/pull/229

Issue #87, opened 2020-05-31
Restore allocate_varlen_buffer 


pearu:
The usage of `allocate_varlen_buffer` was replaced with `calloc` in PR #84 as a temporary solution to the issue #75. However, not being able to use functions defined by Omniscidb server blocks several planned features: cuda enabled UDFs, calling library functions (that are linked to the server) from UDFs, etc. So, we'll need to restore the possibility to use `allocate_varlen_buffer` or any other low-level function from rbc generated LLVM IR strings.

A few items to investigate:
1) Do the order of declared functions matter on LLVM IR?
2) Is the function `allocate_varlen_buffer` being declared more than once?
3) Find the usage of `locals` on `compiler.compile_extra`

_Originally posted by @guilhermeleobas in https://github.com/xnd-project/rbc/issues/75#issuecomment-635535014 and edited by @pearu_

guilhermeleobas:
Steps to reproduce:

1. Apply this diff
```diff
diff --git a/rbc/omnisci_array.py b/rbc/omnisci_array.py
index 7ea7c66..cb121f0 100644
--- a/rbc/omnisci_array.py
+++ b/rbc/omnisci_array.py
@@ -69,7 +69,7 @@ def omnisci_array_constructor(context, builder, sig, args):
     alloc_fnty = ir.FunctionType(int8_t.as_pointer(), [int64_t, int64_t])
     # see https://github.com/xnd-project/rbc/issues/75
     alloc_fn = builder.module.get_or_insert_function(
-        alloc_fnty, name="calloc")
+        alloc_fnty, name="allocate_varlen_buffer")
     ptr8 = builder.call(alloc_fn, [element_count, element_size])
     builder_buffers[builder].append(ptr8)
     ptr = builder.bitcast(ptr8, context.get_value_type(ptr_type))
```

2. Run the following test:
```
pytest -sv rbc/tests/test_omnisci_array.py::test_np_cumsum -x --capture=no --disable-warnings
```

guilhermeleobas:
I was able to reproduce this error on Numba. I think it will be easier to solve the problem from this example:
https://gist.github.com/guilhermeleobas/a5e7204f325b526bfe7f01895e0cbc68

guilhermeleobas:
xref https://github.com/numba/numba/issues/6012

guilhermeleobas:
I was finally able to compile a code using `allocate_varlen_buffer`. For that, I had to change a few internals:

RBC:
[...]

PR #236, opened 2020-12-18
Restore allocate_varlen_buffer

guilhermeleobas:
Solves gh-87

In this PR we're doing two things:
1. Remove target_info argument from RemoteJIT* classes
2. Hackish way to prevent numba from calling _ensure_finalize
3. Replace "calloc" by "allocate_varlen_buffer"

~PR is currently segfaulting in test `test_omnisci_math::positive`:~

```
~pytest rbc/tests/test_omnisci_math.py -k 'positive'~
```

cc @pearu

### DIFF
--- a/rbc/irtools.py
+++ b/rbc/irtools.py
@@ -3,6 +3,7 @@
 
 import re
 import warnings
+from contextlib import contextmanager
 from collections import defaultdict
 from llvmlite import ir
 import llvmlite.binding as llvm
@@ -207,6 +208,15 @@ class JITRemoteGPUTypingContext(typing.Context):
 # Code generation methods
 
 
+@contextmanager
+def replace_numba_internals_hack():
+    # Hackish solution to prevent numba from calling _ensure_finalize. See issue #87
+    _internal_codegen_bkp = registry.cpu_target.target_context._internal_codegen
+    registry.cpu_target.target_context._internal_codegen = JITRemoteCPUCodegen("numba.exec")
+    yield
+    registry.cpu_target.target_context._internal_codegen = _internal_codegen_bkp
+
+
 def make_wrapper(fname, atypes, rtype, cres, target: TargetInfo, verbose=False):
     """Make wrapper function to numba compile result.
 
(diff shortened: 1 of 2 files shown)


guilhermeleobas:
Circle CI failure is not related to any changes introduced in this PR

rbc/irtools.py
```diff
@@ -140,18 +167,17 @@ def init(self):
...
 
-    def __init__(self, typing_context, target_info):
-        self.target_info = target_info
+    def __init__(self, typing_context):
         super().__init__(typing_context)
 
     @compiler_lock.global_compiler_lock
     def init(self):
-        self.address_size = self.target_info.bits
+        target_info = TargetInfo()
```
guilhermeleobas, line 177: I guess this is not required anymore

rbc/irtools.py
```diff
@@ -104,25 +125,31 @@ def _get_host_cpu_features(self):
...
-    def __init__(self, typing_context, target_info):
-        self.target_info = target_info
-        super(RemoteCPUContext, self).__init__(typing_context)
+    def __init__(self, typing_context):
+        super(JITRemoteCPUContext, self).__init__(typing_context)
 
     @compiler_lock.global_compiler_lock
     def init(self):
-        self.address_size = self.target_info.bits
+        target_info = TargetInfo()
```
guilhermeleobas, line 146: same

rbc/irtools.py
```diff
@@ -356,84 +375,85 @@ def compile_to_LLVM(functions_and_signatures,
...
+        if target_info.is_cpu:
             typing_context = typing.Context()
-            target_context = RemoteCPUContext(typing_context, target)
-        elif target.is_gpu:
-            typing_context = RemoteGPUTypingContext()
-            target_context = RemoteGPUTargetContext(typing_context, target)
+            target_context = JITRemoteCPUContext(typing_context)
+        elif target_info.is_gpu:
+            typing_context = JITRemoteGPUTypingContext()
+            target_context = JITRemoteGPUTargetContext(typing_context)
```
pearu, line 390: It looks like we can eliminate `target_info.is_gpu` block here and merge `JITRemoteCPUContext` and `JITRemoteGPUTargetContext` to `JITRemoteTargetContext` to simplify irtools.py code a little.
After all, for GPU LLVM IR we are using numba CPU code generation with minor adjustments.

rbc/irtools.py
pearu: I am not sure what do you mean but I'd keep `is32bit` and `address_size` to correspond to the bitness of the target: some GPU cards are 32-bit while the server runs 64-bit linux.

pearu (approved):
LGTM. I rebased the PR with master and did some code clean up.

I am fine with merging this PR as it is but I believe that irtools.py could be simplified a little, see my comments.

rbc/irtools.py
guilhermeleobas: Actually, we need the `if-else` block to map math functions to libdevice when the target has GPU enabled. See:
https://github.com/xnd-project/rbc/blob/a34604c848250b670a79b214effffb8546ccdc7a/rbc/irtools.py#L182-L189

guilhermeleobas:
>I am not sure what do you mean but I'd keep is32bit and address_size to correspond to the bitness of the target: some GPU cards are 32-bit while the server runs 64-bit linux.

It is still present, I just removed `target_info` from the list of arguments. See:
https://github.com/xnd-project/rbc/blob/a34604c848250b670a79b214effffb8546ccdc7a/rbc/irtools.py#L160-L163

PR #239, opened 2020-12-29
Support Column is_null (replaces PR #230)

### LINKED ISSUE #188: Null check on column accessor

pearu:
Something went wrong with rebase, so created a new PR for #230 .

Fix #188

Requires: omnisci/omniscidb-internal#5104 (was omnisci/omniscidb-internal#5074)

Edit: test passes on both cpu and gpu versions of omniscidb.


rbc/tests/test_omnisci_column_basic.py
```diff
@@ -919,3 +919,49 @@ def convolve(x, kernel, m, y):
...
+
+    @omnisci(f'int32(Column<{typ}>, RowMultiplier, OutputColumn<{typ}>)')
+    def my_row_copier_mul(x, m, y):
+        input_row_count = len(x)
+        for i in range(input_row_count):
+            if x.is_null(i):
+                y.set_null(i)
+            else:
+                y[i] = x[i] + x[i]
+        return m * input_row_count
```
guilhermeleobas, line 960: should we move this test to `test_omnisci_column_null.py

pearu: actually, this is a duplicate of the test in test_omnisci_column_null.py. I suggest removing the test here.

rbc/omnisci_backend/omnisci_buffer.py
```diff
@@ -347,19 +349,94 @@ def omnisci_buffer_setitem(a, i, v):
 
 
 @extending.intrinsic
-def omnisci_buffer_is_null_(typingctx, data):
-    sig = types.int8(data)
+def omnisci_column_set_null_(typingctx, col_var, row_idx):
```
pearu, line 352: This is Column specific implementation and should be implemented in `omnisci_column.py`

rbc/typesystem.py
```diff
@@ -1212,6 +1222,9 @@ def _demangle(s):
 if nb is not None:
     class Boolean1(nb.types.Boolean):
 
+        def __name__(self):
+            return 'Boolean1'
+
```
pearu, line 1227: Which part of the PR uses `__name__` method? Also, `__name__` is usually an attribute or property, not a method.
Remove if not needed. The same applies to `Boolean8` below.

rbc/omnisci_backend/omnisci_buffer.py
```diff
@@ -347,19 +349,94 @@ def omnisci_buffer_setitem(a, i, v):
...
+        if isinstance(T, types.Float):
+            res = builder.bitcast(res, nv.type)
+
+        return builder.icmp_signed('==', res, nv)
+
+    return sig, codegen
+
+
+@extending.overload_method(BufferType, 'isNull')
+@extending.overload_method(BufferType, 'is_null')
```
pearu, line 418: We should decide which method to support, `isNull` or `is_null`. Not both. See also https://github.com/xnd-project/rbc/issues/240

pearu (commented):
Column specific code should be implemented in omnisci_column.py

rbc/omnisci_backend/omnisci_buffer.py
guilhermeleobas: I think C++ `UDTFs` uses `isNull` but I think it makes sense for us to continue using `is_null` to keep consistent with array.

rbc/typesystem.py
guilhermeleobas: They're were required before when I was constructing a type on `omniscidb.py` but I guess it can be removed now

pearu: Yes, let's remove it if not needed.

rbc/omnisci_backend/omnisci_buffer.py
pearu: Ok, let's use Pythonic `is_null` and remove `isNull` support.

pearu:
LGTM.
When https://github.com/omnisci/omniscidb-internal/pull/5104 is merged, this PR can be merged as well.

Btw, after #241 is merged, this PR may require an update but that would be a good opportunity to resolve also https://github.com/xnd-project/rbc/issues/240 within this PR.

Issue #220, opened 2020-11-22
Support signature templates


pearu:
Current situation:
```
@omnisci('double(double[], int64)', 'float(float[], int64)', 'int64(int64[], int64)', 'int32(int32[], int64)')
def mygetitem(x, i):
    return x[i]
```

Desired situation:
```
@omnisci('T(T[], int64)', T=['double', 'float', 'int64', 'int32'])
def mygetitem(x, i):
    return x[i]
```
or more generally,
```
@omnisci('T(T[], integer, fp, int64)',
         T=<list of numpy|numba|typesystem types>,
         integer=['int8', 'int16', 'int32', 'int64'],
         fp=['float32', 'float64'])
...
```
that will generate `2 * 4 * len(T)` signatures.

pearu:
This issue is a follow-up to #182 .

guilhermeleobas:
I believe this issue is already solved. cc @pearu

pearu:
Yes, closing it.

PR #241, opened 2021-01-02
Implement templates support and make custom type a fundamental type of RBC typesystem

pearu:
As in title.

Having custom type as a fundamental type of typesystem makes introducing new types much easier: one has to derive a custom type from `Type` and implement the `tonumba` method as a minimum.

Resolves issue #220.

### DIFF
--- a/rbc/omnisci_backend/omnisci_column.py
+++ b/rbc/omnisci_backend/omnisci_column.py
@@ -4,40 +4,25 @@ Omnisci Column type is the type of input/output column arguments in
 UDTFs.
 """
 
-__all__ = ['ColumnPointer', 'OutputColumn', 'Column',
-           'OmnisciOutputColumnType', 'OmnisciColumnType',
-           'OmnisciCursorType', 'cursor_type_converter',
-           'output_column_type_converter', 'column_type_converter',
-           'table_function_sizer_type_converter']
+__all__ = ['OutputColumn', 'Column', 'OmnisciOutputColumnType', 'OmnisciColumnType',
+           'OmnisciCursorType']
 
 from rbc import typesystem
-from rbc.utils import get_version
-from .omnisci_buffer import (
-    BufferPointer, Buffer, BufferPointerModel,
-    buffer_type_converter, OmnisciBufferType,
-)
-
-if get_version('numba') >= (0, 49):
-    from numba.core import datamodel
-else:
-    from numba import datamodel
+from .omnisci_buffer import Buffer, OmnisciBufferType
 
 
 class OmnisciColumnType(OmnisciBufferType):
     """Omnisci Column type for RBC typesystem.
     """
+    pass_by_value = True
 
 
-class OmnisciOutputColumnType(OmnisciBufferType):
+class OmnisciOutputColumnType(OmnisciColumnType):
     """Omnisci OutputColumn type for RBC typesystem.
 
     OutputColumn is the same as Column but introduced to distinguish
     the input and output arguments of UDTFs.
     """
-
-
-class ColumnPointer(BufferPointer):
-    """Type class for pointers to :code:`Omnisci Column<T>` structure."""
 
 
 class Column(Buffer):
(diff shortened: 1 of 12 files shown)


rbc/omnisci_backend/omnisci_column.py
```diff
@@ -48,77 +33,27 @@ class OutputColumn(Column):
...
+        for p in args[0]:
+            if not isinstance(p, OmnisciColumnType):
+                p = OmnisciColumnType((p,))
+            params.append(p)
+        return (tuple(params),)
+
     @property
     def as_consumed_args(self):
-        return list(self)
+        return self[0]
```
guilhermeleobas, line 50: `self[0]` is a tuple of types, right? i.e. `(OmnisciColumnType(...), OmnisciColumnType(...)`. Can you add a type annotation to it?





```suggestion
@property
def as_consumed_args(self) -> Tuple[typesystem.Type, ...]:
    return self[0]
```

rbc/remotejit.py
```diff
@@ -65,6 +104,7 @@ def debug(self):
     def local(self):
         sig = Signature(self.remotejit.local)
         sig.signatures.extend(self.signatures)
+        # sig.signature_templates.update(self.signature_templates)
```
guilhermeleobas, line 107: Remove

rbc/remotejit.py
```diff
@@ -185,21 +240,32 @@ def normalized(self, func):
                     print(f'{type(self).__name__}.normalized: skipping {sig} as'
                           f' not supported by devices: {devices}')
                 continue
+            templates = self.signature_templates.get(sig, {})
             sig = Type.fromobject(sig)
             if not sig.is_complete:
+                print('Ignore incomplete type: ', sig)
```
guilhermeleobas, line 246: Either remove or replace `print` by `warnings.warn`

rbc/tests/test_omnisci_column_arguments.py
```diff
@@ -81,7 +81,7 @@ def text_rbc_copy_cccc_rowmul(x, x2, x3, x4, m, y, y2, y3, y4):
                           'cursor(i4,i8);f8', 'cursor(i4,i8);cursor(f8,f4)',
                           'cursor(i4,i8,f8,f4)'])
 def test_copy(omnisci, inputs):
-    omnisci.require_version((5, 6), 'Requires omniscidb-internal PR 5134')
+    omnisci.require_version((5, 7), 'Requires omniscidb-internal PR 5134')
```
guilhermeleobas, line 84: This is to avoid pytest to execute this test?  If so, perhaps is it better to skip the test with `@pytest.mark.skip(reason)` decorator and remove it after a new version of omniscidb is released.

guilhermeleobas (approved):
Looks good to me. Just some comments

rbc/omnisci_backend/omnisci_column.py
guilhermeleobas: You do not need to do this in this PR but at some point we should follow numba and start annotating the codebase

pearu: I agree we should use annotation to make code easier to read.
However, this PR implements nontrivial features, and to keep diff size of relevant implementation detail small, let's introduce annotations in separate PRs with no new features.

rbc/remotejit.py
pearu: Fixed via `assert not self.signature_templates` - `local` is used mostly for debugging and when this is needed in relation to templates then we should add the supporting code here.

rbc/remotejit.py
pearu: Using `warn` as a fix.

rbc/tests/test_omnisci_column_arguments.py
pearu: https://github.com/omnisci/omniscidb-internal/pull/5134 has been merged. So, this test can be enabled again.

PR #243, opened 2021-01-04
Replace miniconda -> minimamba on circleci

guilhermeleobas:
CircleCI was failing due to miniconda using more memory than what it was available. Replacing miniconda by minimamba seems to solve it.


pearu (approved):
LGTM

PR #244, opened 2021-01-07
null support in UDF

### LINKED ISSUE #240: NULL support in UDFs
Currently, UDF Array arguments support `is_null` method for null checking. However, the null checking support for other argument types needs to be verified with tests and the corresponding implementation if needed.

To resolve this issue, test the following UDFs:

```
@omnisci('T(T, T)')
def null_to_value(x, value):
    if x.isNull():  # or `is_null(x)`
        return value
    return x

@omnisci('T(T, T)')
def value_to_null(x, value):
    if x == value:
        return x.null_value  # or `null_value(x)`
    return x
```

using test table with nulls (see rbc/tests/__init__.py).

The tests must be created for all `T` in `int8, ..., int64, bool, float, double, Array<int8>, ..., Array<double>, Bytes`.

guilhermeleobas:
Fix #240

### DIFF
--- a/rbc/omnisci_backend/omnisci_buffer.py
+++ b/rbc/omnisci_backend/omnisci_buffer.py
@@ -389,6 +389,18 @@ def omnisci_buffer_is_null_(typingctx, data):
 
 
 @extending.intrinsic
+def omnisci_buffer_set_null_(typingctx, data):
+    sig = types.none(data)
+
+    def codegen(context, builder, sig, args):
+        rawptr = cgutils.alloca_once_value(builder, value=args[0])
+        ptr = builder.load(rawptr)
+        builder.store(int8_t(1), builder.gep(ptr, [int32_t(0), int32_t(2)]))
+
+    return sig, codegen
+
+
+@extending.intrinsic
 def omnisci_array_is_null_(typingctx, T, elem):
     sig = types.boolean(T, elem)
 
@@ -418,4 +430,46 @@ def omnisci_buffer_is_null(x, row_idx=None):
         else:
             def impl(x, row_idx=None):
                 return omnisci_array_is_null_(T, x[row_idx])
+        return impl
+
+
+@extending.intrinsic
+def omnisci_array_set_null_(typingctx, arr, row_idx):
+    T = arr.eltype
+    sig = types.none(arr, row_idx)
+
+    target_info = TargetInfo()
+    null_value = target_info.null_values[f'Array<{T}>']
+    # it looks like that the null values for arrays are:
+    null_value = (null_value - 1) * (-1)
+
+    def codegen(context, builder, signature, args):
+        # get the operator.setitem intrinsic
+        fnop = context.typing_context.resolve_value_type(omnisci_buffer_ptr_setitem_)
+        setitem_sig = types.none(arr, row_idx, T)
+        # register the intrinsic in the typing ctx
+        fnop.get_call_type(context, setitem_sig.args, {})
+        intrinsic = context.get_function(fnop, setitem_sig)
+
+        data, index = args
+        # data = {T*, i64, i8}*
+        ty = data.type.pointee.elements[0].pointee
+        nv = ir.Constant(ir.IntType(T.bitwidth), null_value)
+        if isinstance(T, types.Float):
+            nv = builder.bitcast(nv, ty)
+        intrinsic(builder, (data, index, nv,))
+
+    return sig, codegen
+
+
+@extending.overload_method(BufferPointer, 'set_null')
+def omnisci_buffer_set_null(x, row_idx=None):
+    if isinstance(x, BufferPointer):
+        if cgutils.is_nonelike(row_idx):
+            def impl(x, row_idx=None):
+                return omnisci_buffer_set_null_(x)
+        else:
+            def impl(x, row_idx=None):
+                return omnisci_array_set_null_(x, row_idx)
+            return impl
         return impl
--- a/rbc/omnisci_backend/omnisci_buffer.py
+++ b/rbc/omnisci_backend/omnisci_buffer.py
@@ -26,6 +26,7 @@ Omnisci buffer objects from UDF/UDTFs.
 import operator
 from collections import defaultdict
 from llvmlite import ir
+import numpy as np
 from rbc import typesystem
 from rbc.utils import get_version
 from rbc.targetinfo import TargetInfo
@@ -183,7 +184,12 @@ def omnisci_buffer_constructor(context, builder, sig, args):
     fa.ptr = ptr                  # T*
     fa.sz = element_count         # size_t
     if null_type is not None:
-        is_null = context.get_value_type(null_type)(0)
+        is_zero = builder.icmp_signed('==', element_count, int64_t(0))
+        with builder.if_else(is_zero) as (then, orelse):
+            with then:
+                is_null = context.get_value_type(null_type)(1)
+            with orelse:
+                is_null = context.get_value_type(null_type)(0)
         fa.is_null = is_null      # int8_t
     return fa._getpointer()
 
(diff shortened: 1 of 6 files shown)


guilhermeleobas:
There's something wrong with NULL values inside arrays. See:

xref: https://github.com/xnd-project/rbc/issues/245#issuecomment-756352346

rbc/omnisci_backend/omnisci_buffer.py
```diff
@@ -387,12 +388,88 @@ def codegen(context, builder, sig, args):
...
+        intrinsic = context.get_function(fnop, setitem_sig)
+
+        data, index = args
+        # data = {T*, i64, i8}*
+        ty = data.type.pointee.elements[0].pointee
+        nv = ir.Constant(ir.IntType(T.bitwidth), null_value)
+        if isinstance(T, types.Float):
+            nv = builder.bitcast(nv, ty)
+        intrinsic(builder, (data, index, nv,))
+
```
guilhermeleobas, line 461: rather than copying the setitem implementation, this is calling the `operator.setitem` intrinsic for arrays.

rbc/omnisci_backend/omnisci_buffer.py
```diff
@@ -387,12 +388,88 @@ def codegen(context, builder, sig, args):
...
+@extending.intrinsic
+def omnisci_array_set_null_(typingctx, arr, row_idx):
+    T = arr.eltype
+    sig = types.none(arr, row_idx)
+
+    target_info = TargetInfo()
+    null_value = target_info.null_values[f'Array<{T}>']
+    # it looks like that the null values for arrays are:
+    null_value = (null_value - 1) * (-1)
```
pearu, line 444: Notice that omnisci sends the null values as unsigned integers. Hence, omnisci_array_is_null_ should use unsigned version of `icmp`.

pearu: Another note: the null_value (in Python) is misinterpreted as a signed integer.
Using Python int (that `null_value` is here), the expression `(null_value - 1) * (-1)` is valid (because Python int has no bitwidth) but when using fixed width integer (say, numpy int dtype), it would be integer overflow.

I hope we can bring some sense to this transformation before accepting it.

pearu:
```
>>> null_value = 9223372036854775809 # 'Array<int64>'
>>> str(ir.Constant(ir.IntType(64), numpy.uint64(null_value).astype(numpy.int64)))
'i64 -9223372036854775807'
>>> str(ir.Constant(ir.IntType(64), null_value))
'i64 9223372036854775809'
>>> str(ir.Constant(ir.IntType(64), -(null_value-1)))
'i64 -9223372036854775808'
```
I would expect `i64 -9223372036854775807` to be correct rather than `i64 -9223372036854775808`.

guilhermeleobas: I figure it out what was wrong. The server is sending null values as unsigned ones rather than signed. See:
https://godbolt.org/z/37dnvd

guilhermeleobas:
Pull request is ready for review. In the last commits I've included the following:
- `array.is_null()`
- `array.is_null(index)`
- `array.set_null(index)`
- `np.array(other_array)` - for copying an array
- `np.empty_like(an_empty_array)` - to create an empty array
- Set `array.is_null` struct member to `True` when size in `array(size, dtype)` is 0

rbc/omnisci_backend/omnisci_buffer.py
pearu: So,
```
>>> null_value = 9223372036854775809 # 'Array<int64>'
>>> str(ir.Constant(ir.IntType(64), numpy.uint64(null_value).astype(numpy.int64)))
'i64 -9223372036854775807'
```
corresponds to the result that you referred in https://godbolt.org/z/37dnvd , as expected.
Yes, the server sends the null values as (stringified) unsigned ints (which is not wrong, it is just a choice).
When `ir.Constant` is transformed to a LLVM IR string, we should see the null value as the corresponding signed int.
So I think the correct way to create the `ir.Constant` instance is to use:
```
  ir.Constant(ir.IntType(64), numpy.uint64(null_value).astype(numpy.int64)
```

guilhermeleobas: Except for the last part, that's what I did:
```python
T = ... # typeof null_value
nv = numpy.uint64(null_value).astype(T)
constant = ir.Constant(ir.IntType(bits), nv)
```

See:
https://github.com/xnd-project/rbc/blob/ba5a169c97b0cf8baa4cb176fd45700b033b992d/rbc/omnisci_backend/omnisci_buffer.py#L451-L456

pearu: Shouldn't the same transform be applied to float types? E.g. `null_value = np.uint64(null_value).astype(np.int64)` for float64. Also, np.uint64(null_value) does not sound right, say, for int16 null value because `np.uint64(..).astype('int16')` looses bits. I think it should read `np.uint16(..).astype(np.int16)` for int16 null value, for instance.

Or better yet, have:
```
null_value = np.uint64(null_value)
nv = ir.Constant(ir.IntType(T.dtype.bitwidth), null_value)
# in is_null codegen:
return builder.icmp_unsigned('==', elem, nv)
# set_null code needs to be tested
```

The aim is to obtain code with fewer conditionals and fewer uint->int transformations as we just want the bit content of null values to be correct.

guilhermeleobas: > Shouldn't the same transform be applied to float types?

No. For floats the server sends the correct value. See: https://godbolt.org/z/x4of5o

> Also, np.uint64(null_value) does not sound right, say, for int16 null value because np.uint64(..).astype('int16') looses bits. I think it should read np.uint16(..).astype(np.int16) for int16 null value, for instance.

Well, I tested all the combinations on my machine and `np.uint64` seems to be converting the data correctly. Nevertheless, I can change this to `np.uT(nv).astype(T)` without any problems.

> `nv = ir.Constant(ir.IntType(T.dtype.bitwidth), null_value)`

This is not safe. llvmlite accepts any value for null_value and it will not overflow it as we want. For instance,
`ir.Constant(ir.IntType(8), 129) != -127`

pearu: > > Shouldn't the same transform be applied to float types?
> 
> No. For floats the server sends the correct value. See: https://godbolt.org/z/x4of5o

This correctness test relies on the used printf formatting. Notice that you are printing unsigned values using signed formatting and hence the negative values.

> > Also, np.uint64(null_value) does not sound right, say, for int16 null value because np.uint64(..).astype('int16') looses bits. I think it should read np.uint16(..).astype(np.int16) for int16 null value, for instance.
> 
> Well, I tested all the combinations on my machine and `np.uint64` seems to be converting the data correctly. Nevertheless, I can change this to `np.uT(nv).astype(T)` without any problems.

It will likely matter when endianness becomes a factor.

> 
> > `nv = ir.Constant(ir.IntType(T.dtype.bitwidth), null_value)`
> 
> This is not safe. llvmlite accepts any value for null_value and it will not overflow it as we want. For instance,
> `ir.Constant(ir.IntType(8), 129) != -127`

Makes sense, although, I would not want to deal with overflow, just having the memory bits correct is sufficient..

pearu (approved):
LGTM

pearu (commented):
It looks like https://github.com/xnd-project/rbc/issues/254 needs to be resolved.

guilhermeleobas:
> It looks like #254 needs to be resolved.

Done!

guilhermeleobas:
CI seems to be stuck on github interface but it's all green on travis CI

pearu (approved):
LGTM, thanks @guilhermeleobas !

PR #256, opened 2021-01-19
Add Windows build to CircleCI

guilhermeleobas:
As title.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,46 +1,95 @@
-version: 2
+version: 2.1
+
+orbs:
+  docker: circleci/docker@1.5.0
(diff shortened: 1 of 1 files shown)


guilhermeleobas:
My idea was to run a ubuntu container inside the windows VM to spin OmniSciDB. Unfortunately, there are a couple of limitations with running docker inside a circleCI "machine" which prevents the host from communicating with the container. See:
https://discuss.circleci.com/t/circleci-config-yml-with-more-than-one-docker-image/24370/10

There is a workaround which is to use `machine` but circleCI docs say they plan to charge for this "feature" in a near future.

PR #257, opened 2021-01-20
test with gh-actions

guilhermeleobas:


guilhermeleobas:
I guess we can drop Circle CI and appveyor if this PR gets approved. This PR creates a matrix for:
- Python 3.6, 3.7 and 3.8 (numba doesn't support 3.9 yet)
- Omniscidb 5.2, 5.4 and 5.5 (ubuntu only)
- ubuntu, windows and mac

It takes 10 minutes to run all combinations and it only executes if flake8 succeed.

pearu (approved):
LGTM

PR #268, opened 2021-01-29
Add command to declare external functions

### LINKED ISSUE #263: Expose external functions as callables
The goal is to expose external functions using a friendly API (i.e. a decorator) which internally will insert function declaration in the LLVM module.

One way I though about doing this was to introduce `external` keyword to `@omnisci` which declares if the function is an external or not. For instance:

```python
@omnisci('T(T)', T=['float64'], external=True)
def log2(x):
    pass

@omnisci('T(T)', T=['float64'])
def fn(x):
    return log2(x)
```

guilhermeleobas:
Closes #263 

Quick note: `cmath.remainder` returns negative value in some cases. I'll investigate it further in a couple of days

### DIFF
--- a/rbc/external.py
+++ b/rbc/external.py
@@ -4,10 +4,35 @@ from numba.core import extending, funcdesc, types
 from rbc.typesystem import Type
 
 
-class _External:
-    def __init__(self, symbol, signature):
-        self.symbol = symbol
+class External:
+    @classmethod
+    def fromobject(cls, signature, name=None):
+        # Make inner function for the actual work
+        t = Type.fromobject(signature)
+        if not t.is_function:
+            raise ValueError("signature must represent a function type")
+
+        if name is None:
+            name = t.name
+        if not name:
+            raise ValueError(
+                f"external function name not specified for signature {signature}"
+            )
+
+        return cls(name, t.tonumba())
+
+    def __init__(self, name: str, signature: types.FunctionType):
+        """
+        Parameters
+        ----------
+        name : str
+            The name of the external function
+        signature : Numba function signature
+            A numba function type signature. i.e. (float64, float64) -> int64
+        """
         self._signature = signature
+        self.name = name
+        self.register()
 
     def register(self):
         # typing
@@ -18,7 +43,7 @@ class _External:
         )
 
         template = make_concrete_template(
-            self.symbol, key=self.symbol, signatures=[self.signature]
+            self.name, key=self.name, signatures=[self.signature]
         )
         infer(template)
         infer_global(self, types.Function(template))
@@ -26,18 +51,18 @@ class _External:
         # lowering
         def codegen(context, builder, sig, args):
             fndesc = funcdesc.ExternalFunctionDescriptor(
-                self.symbol, sig.return_type, sig.args
+                self.name, sig.return_type, sig.args
             )
             func = context.declare_external_function(builder.module, fndesc)
             return builder.call(func, args)
 
-        extending.lower_builtin(self.symbol, *self.signature.args)(codegen)
+        extending.lower_builtin(self.name, *self.signature.args)(codegen)
 
     def __call__(self, *args, **kwargs):
         """
         This is only defined to pretend to be a callable from CPython.
         """
-        msg = f"{self.symbol} is not usable in pure-python"
+        msg = f"{self.name} is not usable in pure-python"
         raise NotImplementedError(msg)
 
     @property
@@ -52,25 +77,5 @@ class _External:
     def signature(self):
         return self._signature
 
-    @property
-    def name(self):
-        return self.symbol
 
-
-def external(signature, name=None):
-
-    # Make inner function for the actual work
-    t = Type.fromobject(signature)
-    if not t.is_function:
-        raise ValueError("signature must represent a function type")
-
-    if name is None:
-        name = t.name
-    if not name:
-        raise ValueError(
-            f"external function name not specified for signature {signature}"
-        )
-
-    llc = _External(name, t.tonumba())
-    llc.register()
-    return llc
+external = External.fromobject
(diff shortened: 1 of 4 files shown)


rbc/external.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def _lower_external_call(self):
+        def codegen(context, builder, sig, args):
+            fndesc = funcdesc.ExternalFunctionDescriptor(
+                self.symbol, sig.return_type, sig.args
+            )
+            func = context.declare_external_function(builder.module, fndesc)
+            return builder.call(func, args)
+
+        extending.lower_builtin(self.symbol, *self.signature.args)(codegen)
```
pearu, line 33: Are these methods used by numba? If not, merge the code into `register` method because `_type_infer` and `_lower_external_call` would represent numba internal API that should be irrelevant for rbc.

rbc/typesystem.py
```diff
@@ -493,6 +493,10 @@ def is_function(self):
             isinstance(self[1], tuple) and not isinstance(self[1], Type) \
             and all([isinstance(p, Type) for p in self[1]])
 
+    @property
+    def is_named_function(self):
+        return self.is_function and 'name' in self[0]._params
+
```
pearu, line 499: Delete and rebase.

rbc/external.py
```diff
@@ -0,0 +1,60 @@
+# This code has heavily inspired in the numba.extending.intrisic code
+
+from numba.core import extending, funcdesc, types
+from rbc.typesystem import Type
+
+
+class _External:
```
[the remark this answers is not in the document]
pearu: In order to keep the constructor processing-free, use
```
class External:
    @classmethod
    def fromobject(cls, signature, name=None):
        ...<see above>
        return cls(name, signature)
    def __init__(self, symbol, signature):
        self.symbol = symbol
        self.signature = signature

external = External.fromobject
```

pearu (changes requested):
Thanks, @guilhermeleobas, this PR implements a very useful feature.

I had several suggestions for simplifying the code. Read them as the former comments have precedence over the latter comments.

The external feature provides the third method of calling math functions from UDF/UDTFs (the first and second are based on exposing math and numpy functions) when applied to scalars. So I wonder if these methods have any performance differences and what should be recommended to users?

rbc/external.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def __call__(self, *args, **kwargs):
+        """
+        This is only defined to pretend to be a callable from CPython.
+        """
+        msg = f'{self.symbol} is not usable in pure-python'
+        raise NotImplementedError(msg)
+
+
+def external(*args):
```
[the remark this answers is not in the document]
guilhermeleobas: > Should external support multiple arguments? What would be the use cases?

Hum...I don't think there's any real case for supporting multiple arguments. Consider for instance the [C math library](https://www.cplusplus.com/reference/cmath/acos/), each function supports only a set of input signature. For example, `double sin(double)`, `float sinf(float)`.

> Or better yet:
> 
> ```
> def external(signature, name=None):
> ```
> 
> where signature can be anything convertible to a typesystem function type. When signature is, say, a numba type, then one must provide the name explicitly.

Agree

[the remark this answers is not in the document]
guilhermeleobas: > Should external support multiple arguments? What would be the use cases?

On the other hand, defining a function with a return type of `int64` and calling it with an `int32` ought to work

rbc/tests/test_external.py
```diff
@@ -0,0 +1,224 @@
...
+        pytest.skip(f"cmath function {fname} not supported")
+
+    if fname in ["frexp", "modf", "nan"]:
+        pytest.skip(f"cmath function {fname} crashes omniscidb server")
+
+    if fname in ["remainder"]:
+        pytest.skip(f"cmath.{fname} wrong output!")
+
+    table = omnisci.table_name
+    cmath_func = "omnisci_" + fname
```
pearu, line 128: One could also use
```
cmath_func = table + "_" + fname
```
to ensure test udf naming locality.

rbc/external.py
```diff
@@ -0,0 +1,81 @@
...
+    def register(self):
+        # typing
+        from numba.core.typing.templates import (
+            make_concrete_template,
+            infer_global,
+            infer,
+        )
+
+        template = make_concrete_template(
+            self.name, key=self.name, signatures=[self.signature]
```
pearu, line 46: What happens when the `(self.name, self.signature)` pair has already registered? Would it be possible to issue a warning?

rbc/tests/test_external.py
```diff
@@ -0,0 +1,224 @@
...
+def omnisci():
+
+    for o in omnisci_fixture(globals()):
+        define(o)
+        yield o
+
+
+cmath = (
+    # Trigonometric
+    ("cos", "float64(float64)"),
```
pearu, line 19: Someday, I'd like to move these `(name, signature)` pairs to libfuncs.py.  It is not relevant for this PR though.

rbc/external.py
```diff
@@ -0,0 +1,81 @@
+# This code has heavily inspired in the numba.extending.intrisic code
+
+from numba.core import extending, funcdesc, types
+from rbc.typesystem import Type
+
+
+class External:
+    @classmethod
+    def fromobject(cls, signature, name=None):
+        # Make inner function for the actual work
```
pearu, line 10: Could you add a docstring here as this method is exposed to users?

pearu (approved):
LGTM. I approved it but could you also add a docstring to `fromobject` method? After all CI a green, go ahead and merge.

rbc/external.py
guilhermeleobas: Former will be replaced by the latter

PR #273, opened 2021-02-05
Change external behaviour to work with hardware-specific types

guilhermeleobas:
As title

### DIFF
--- a/rbc/external.py
+++ b/rbc/external.py
@@ -1,7 +1,8 @@
 # This code has heavily inspired in the numba.extending.intrisic code
 
-from numba.core import extending, funcdesc, types
+from numba.core import extending, funcdesc, types, typing
 from rbc.typesystem import Type
+from rbc.targetinfo import TargetInfo
 
 
 class External:
@@ -16,20 +17,25 @@ class External:
             The name of the external function
         """
         # Make inner function for the actual work
-        t = Type.fromobject(signature)
-        if not t.is_function:
-            raise ValueError("signature must represent a function type")
+        target_info = TargetInfo.host()
+        with target_info:
+            t = Type.fromobject(signature)
+            if not t.is_function:
+                raise ValueError("signature must represent a function type")
 
-        if name is None:
-            name = t.name
-        if not name:
-            raise ValueError(
-                f"external function name not specified for signature {signature}"
-            )
+            if name is None:
+                name = t.name
+            if not name:
+                raise ValueError(
+                    f"external function name not specified for signature {signature}"
+                )
 
-        return cls(name, t.tonumba())
+            typ = t
+            t = t.tonumba()
 
-    def __init__(self, name: str, signature: types.FunctionType):
+        return cls(name, t, typ)
+
+    def __init__(self, name: str, signature: types.FunctionType, typ: Type):
         """
         Parameters
         ----------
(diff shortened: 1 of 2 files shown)


rbc/external.py
```diff
@@ -16,55 +17,64 @@ def fromobject(cls, signature, name: str = None):
...
         """
         # Make inner function for the actual work
-        t = Type.fromobject(signature)
-        if not t.is_function:
-            raise ValueError("signature must represent a function type")
+        target_info = TargetInfo.host()
+        with target_info:
+            t = Type.fromobject(signature)
+            if not t.is_function:
+                raise ValueError("signature must represent a function type")
```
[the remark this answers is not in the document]
pearu: We should merge https://github.com/xnd-project/rbc/pull/274 first, and then use
```
with TargetInfo.dummy():
    if not Type.fromobject(signature).is_function:
        raise ValueError("signature must represent a function type")
```

rbc/external.py
```diff
@@ -16,55 +17,64 @@ def fromobject(cls, signature, name: str = None):
...
+            if name is None:
+                name = t.name
+            if not name:
+                raise ValueError(
+                    f"external function name not specified for signature {signature}"
+                )
 
-        return cls(name, t.tonumba())
+            typ = t
+            t = t.tonumba()
```
pearu, line 34: We should not materialize the signature here!

rbc/external.py
```diff
@@ -16,55 +17,64 @@ def fromobject(cls, signature, name: str = None):
...
         Parameters
         ----------
         name : str
             The name of the external function
         signature : Numba function signature
             A numba function type signature. i.e. (float64, float64) -> int64
+        rbc_type: RBC typesystem Type
+            A type from the RBC typesystem type
         """
         self._signature = signature
```
pearu, line 49: We should eliminate `_signature` that is numba function type.

pearu: And rename `rbc_type` -> `signature`

pearu (changes requested):
Looks good.

`External` properties `return_type`, `args` are not used and should be deleted.

rbc/external.py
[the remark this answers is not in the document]
guilhermeleobas: > This basically checks that signature represents a function

Not just that, but to parse the signature and get the function name if `name` is not specified as a keyword arg

rbc/external.py
```diff
@@ -44,27 +47,28 @@ def __init__(self, name: str, signature: types.FunctionType):
...
+                t = Type.fromobject(self.obj._signature).tonumba()
 
-        # lowering
-        def codegen(context, builder, sig, args):
-            fndesc = funcdesc.ExternalFunctionDescriptor(
-                self.name, sig.return_type, sig.args
-            )
-            func = context.declare_external_function(builder.module, fndesc)
-            return builder.call(func, args)
+                name = self.key
```
pearu, line 57: When external is going to support multiple signatures, as in `external(['double log2(double)', 'float log2f(float)'], name='log2')`, then we should likely use:
```
name = t.name or self.key
```

rbc/external.py
```diff
@@ -44,27 +47,28 @@ def __init__(self, name: str, signature: types.FunctionType):
...
+            obj = self
+            key = self.name
 
-        template = make_concrete_template(
-            self.name, key=self.name, signatures=[self.signature]
-        )
-        infer(template)
-        infer_global(self, types.Function(template))
+            def generic(self, args, kws):
+                t = Type.fromobject(self.obj._signature).tonumba()
```
pearu, line 55: I presume that in multiple signature cases (see below), this is the place where `t` needs to depend on `args`.

pearu (approved):
LGTM.

Added few comments for future extensions.

rbc/external.py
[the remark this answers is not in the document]
pearu: Right. In future, when the signatures need to contain the function names, we may want to eliminate the `name` argument.
So, let's not use `name` too much :)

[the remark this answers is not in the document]
guilhermeleobas: The name might be useful to use as a key of `AbstractTemplate`.

rbc/external.py
guilhermeleobas: exactly

Issue #270, opened 2021-02-03
external decorator should accept overloading


guilhermeleobas:
For instance, the following ought to work:

```python
fn = external(['float sinf(float)', 'double sin(double)'])
```

This can be done by dispatching the correct function call inside codegen

Issue #242, opened 2021-01-04
RuntimeWarning: invalid value encountered in numpy fn


guilhermeleobas:
```
...
rbc/tests/test_omnisci_array_math.py::test_omnisci_array_binary_math[power]
  /home/guilhermel/git/rbc/rbc/tests/test_omnisci_array_math.py:132: RuntimeWarning: invalid value encountered in power
    expected = getattr(np, method)(row_a, row_b)

rbc/tests/test_omnisci_array_math.py::test_omnisci_array_unary_math_fns[log]
  /home/guilhermel/git/rbc/rbc/tests/test_omnisci_array_math.py:279: RuntimeWarning: invalid value encountered in log
    expected = getattr(np, method)(row)

rbc/tests/test_omnisci_array_math.py::test_omnisci_array_unary_math_fns[log2]
  /home/guilhermel/git/rbc/rbc/tests/test_omnisci_array_math.py:279: RuntimeWarning: invalid value encountered in log2
    expected = getattr(np, method)(row)

rbc/tests/test_omnisci_array_math.py::test_omnisci_array_unary_math_fns[log10]
  /home/guilhermel/git/rbc/rbc/tests/test_omnisci_array_math.py:279: RuntimeWarning: invalid value encountered in log10
    expected = getattr(np, method)(row)

rbc/tests/test_omnisci_array_math.py::test_omnisci_array_unary_math_fns[log1p]
  /home/guilhermel/git/rbc/rbc/tests/test_omnisci_array_math.py:279: RuntimeWarning: invalid value encountered in log1p
    expected = getattr(np, method)(row)

rbc/tests/test_omnisci_array_math.py::test_omnisci_array_unary_math_fns[sqrt]
  /home/guilhermel/git/rbc/rbc/tests/test_omnisci_array_math.py:279: RuntimeWarning: invalid value encountered in sqrt
    expected = getattr(np, method)(row)

rbc/tests/test_omnisci_array_math.py::test_omnisci_array_unary_math_fns[arcsin]
  /home/guilhermel/git/rbc/rbc/tests/test_omnisci_array_math.py:279: RuntimeWarning: invalid value encountered in arcsin
    expected = getattr(np, method)(row)

rbc/tests/test_omnisci_array_math.py::test_omnisci_array_unary_math_fns[arccos]
  /home/guilhermel/git/rbc/rbc/tests/test_omnisci_array_math.py:279: RuntimeWarning: invalid value encountered in arccos
[...]

PR #275, opened 2021-02-11
External with device parameter

### LINKED ISSUE #265: Several RBC tests fail when using CUDA enabled omniscidb server
As in title.

guilhermeleobas:
Fixes #265, #270, #242

Current issues:
- [x] `math.ceil` fails and always returns the biggest integer possible when it is mapped to cmath `ceil` or libdevice `__nv_ceil`.
- [x] `math.log2` fails during typing
- [x] `math.remainder` fails during typing
- [x] `math.floor` gives the wrong result
- [x] `math.trunc` gives the wrong result

TO-DO:
- [x] Drop `name` keyword and use the first name as key.
- [x] Avoid calling `register()` in every scenario. For instance, we don't need to do typing for `math.*` functions.
- [x] Introduce a `Registry()` in omnisci_mathimply.py. Currently we are lowering functions using `extending.lower_builtin`.


guilhermeleobas:
This PR introduces and changes the following:
- Changes `External` to dispatch the implementation based on `TargetInfo.is_cpu/is_gpu`
- Introduce `omnisci_mathimpl.py` and `omnisci_mathtyping.py`. The former lowers `math.*` functions using new `External` behavior while the latter adds typing information to `math.log2` and `math.remainder`.
- Remove `name` keyword from `External.fromobject`
- Add two keywords (`typing=True`, `lowering=True`) to `External.fromobject` to indicate if `External` should automatically type/lower or not a given external function call.
- typing and lowering of `omnisci_mathimpl` and `omnisci.mathtyping` are done in separated `Registry()`. Previously I was using the global registry from Numba (`extending.lower_builtin`)

rbc/external.py
```diff
@@ -1,38 +1,65 @@
...
             The name of the external function
+        typing : bool
+            Indicates if External should do typing or not. Default is True
+        lowering: bool
+            Indicates if External should do lowering or not. Default is True. Except for
+            very specific cases, typing and lowering should be True
         """
         # Make inner function for the actual work
         target_info = TargetInfo.dummy()
+        ts = defaultdict(list)
```
pearu, line 27: Notice that all signatures will be with respect to dummy target info. This is problematic because there may exist signatures that contain types that actually depend on the device or on the actual target.

For example, we have `sizeof(long) == 32` on Windows. Eventhough, Type.fromobject does not evaluate `long` when using dummy target, `Type` does not reevaluate the signature when under specific target info (maybe it should but not here).

To avoid any issues that the above might trigger, I suggest storing the original `signature` objects in `ts` rather than the `Type` instances. This will duplicate `Type.fromobject` calls (until we'll support `Type.reevaluate`) but at least the `signature` objects will be transformed to `Type` instances within the correct target info.

rbc/external.py
```diff
@@ -41,9 +68,56 @@ def __init__(self, name: str, signature: types.FunctionType):
...
+        self.typing = typing
+        self.lowering = lowering
+        if self.typing:
+            self.register()
+
+    def __str__(self):
+        a = []
+        for device in self._signatures:
+            for t in self._signatures[device]:
+                a.append(t.tostring())
```
pearu, line 83: nit: use `str(t)` instead of `t.tostring()`

rbc/external.py
```diff
@@ -41,9 +68,56 @@ def __init__(self, name: str, signature: types.FunctionType):
...
+        return ", ".join(a)
+
+    def match_signature(self, atypes):
+        # Code here is the same found in remotejit.py::Signature::best_match
+        device = "CPU" if TargetInfo().is_cpu else "GPU"
+        available_types = self._signatures[device]
```
pearu, line 89: Use
```python
available_types = tuple(map(Type.fromobject, self._signatures[device]))
```
See comments in `External.fromobject` for the context.

rbc/irtools.py
```diff
@@ -134,60 +134,35 @@ def set_env(self, env_name, env):
...
-    def post_lowering(self, mod, library):
-        pass
-
+class JITRemoteTypingContext(typing.Context):
+    def load_additional_registries(self):
+        from rbc.omnisci_backend import omnisci_mathtyping as mathtyping
 
-# ---------------------------------------------------------------------------
-# GPU Context classes
+        self.install_registry(mathtyping.registry)
```
pearu, line 141: This approach needs to be generalized. Namely, users (including the rbc project) may provide `omnisci_mathtyping`-similar modules that should be installed similarly what is in the above, BUT without changing the irtools.py

rbc/omnisci_backend/omnisci_mathtyping.py
```diff
@@ -0,0 +1,28 @@
...
+class Math_unary(ConcreteTemplate):
+    cases = [
+        signature(types.float64, types.int64),
+        signature(types.float64, types.uint64),
+        signature(types.float32, types.float32),
+        signature(types.float64, types.float64),
+    ]
+
+
+if utils.PYVERSION >= (3, 7):
```
pearu, line 21: OT: we may want to consider dropping Python 3.6 support.

pearu (changes requested):
Looks good.

There is one important concern of evaluating signatures against dummy target info. I have added some suggestions on how to postpone the evaluation to use the actual target info(s).

rbc/irtools.py
guilhermeleobas: The same behavior can be achieve using different approaches already available in RBC/Numba: For instance:
- External
- extending.overload_*

rbc/omnisci_backend/omnisci_mathtyping.py
guilhermeleobas: Agreed. Python 3.6 supports ends in 10 months.

rbc/external.py
guilhermeleobas: Done!

rbc/omnisci_backend/__init__.py
```diff
@@ -1,4 +1,5 @@
 from .numpy_funcs import *  # noqa: F401, F403
+from .omnisci_mathimpl import *  # noqa: F401, F403
```
pearu, line 2: IIUC, `omnisci_mathimpl` is not omnisci specific. So, the `omnisci_mathimpl` and `omnisci_mathtyping` should be moved out of `omnisci_backend`. Would it make sense to join these modules into one, say, `rbc/libmath.py` because it mimics what is supported in Python `math` package. 

While supporting Python `math` package from UDF/UDTFs make lots of sense, for the external feature, supporting [cmath](https://en.cppreference.com/w/cpp/header/cmath) makes even more sense as these are the functions that are available when executing the UDF/UDTFs inside the OmniSciDB server.

To make the structure of rbc more clear, I suggest that the external support implementations have dedicated location as follows:
rbc/externals/math.py - provides Python math support as provided in this PR
rbc/externals/cmath.py - provides C math support (not it this PR)
rbc/externals/cstdlb.py - provides C StdLib support ( http://www.cplusplus.com/reference/cstdlib/)
rbc/externals/omniscidb.py - provides OmniSciDB functions declared as `extern "C"` that UDF/UDTFs might need to call.
etc

In this PR, just introduce the `rbc/externals/math.py` module, others will follow in future.

pearu (changes requested):
Thanks, @guilhermeleobas !

See the comments under rbc/omnisci_backend/__init__.py about the suggested structure of supporting external libraries.

All tests pass locally when using CUDA enabled omniscidb. Notice that `test_external` triggers lots of user warnings such as:
```
UserWarning: Skipping test_external_log2__gpu_0 that uses undefined function `log2`
```
Why `omnisci_mathimpl.py` hooks are not effective here?

guilhermeleobas:
> Why omnisci_mathimpl.py hooks are not effective here?

@pearu, `test_external.py` is testing bindings with C math function. I guess what is missing there is to skip the test if GPU is enabled.

rbc/omnisci_backend/__init__.py
guilhermeleobas: > rbc/externals/math.py - provides Python math support as provided in this PR

Notice that this will bind to `math.*` and not `rbc.externals.math.*`.

guilhermeleobas:
Failure is related to a travis CI issue

pearu (approved):
LGTM. Thanks, @guilhermeleobas !

PR #281, opened 2021-02-18
move IS_GPU/IS_CPU out of omniscidb

### LINKED ISSUE #280: IS_CPU/IS_GPU is not specific to omniscidb
The code in
https://github.com/xnd-project/rbc/blob/26af0d176df1d7853510e97d11206144aba5e7ef/rbc/omniscidb.py#L22-L51
should be moved out of omniscidb.py as it is not specific to omniscidb support. Perhaps move to irtools.py that contains other utility functions printf, flush?

guilhermeleobas:
Fix #280


pearu (approved):
LGTM

Issue #284, opened 2021-02-19
Add libdevice bindings using external


guilhermeleobas:
As title

PR #286, opened 2021-02-19
Add cmath and libdevice external functions and gpu bindings to cmath and numpy

### LINKED ISSUE #283: Add cmath bindings using external
As title

guilhermeleobas:
Fixes #283 and #284 

Requires: https://github.com/omnisci/omniscidb-internal/pull/5276

- RBC now uses `sizeof(primitive type)` info send by the server
- Add bindings for the C math library on `rbc/externals/cmath.py`
- Add bindings for libdevice library on `rbc/externals/libdevice.py`
- Add gpu bindinds to math
- Add gpu bindings to numpy
- Update `test_omnisci_math.py` to test functions on both CPU and GPU (when available)


rbc/external.py
```diff
@@ -8,52 +8,67 @@
...
+            targetinfo = TargetInfo.dummy()
+
+        # If we already have a targetinfo registered, then, use an empty context manager
+        try:
+            with targetinfo:
+                pass
+        except AssertionError:
+            import contextlib
+            targetinfo = contextlib.nullcontext()
+
```
guilhermeleobas, line 36: This part of the code is a bit obscure. I wonder if there's a better way to know if a `TargetInfo` object already exists and if it is being used.

We have a few corner cases here:
1. When registering a new registry inside `irtools.py`, a target info was already registered and it is being used
2. When one imports `from rbc.externals import cmath` for instance, a target info probably do not exists

pearu: The fix is here:
https://github.com/xnd-project/rbc/pull/292/commits/3850649f980b8517efefa930a55d4c06a3e2a5bd
that optionally allows nesting target info contexts.

rbc/externals/libdevice.py
```diff
@@ -0,0 +1,40 @@
...
+    class LibDeviceTemplate(typing.templates.AbstractTemplate):
+        key = _key
+
+        def generic(self, args, kws):
+            t = Type.fromstring(f"{retty} {fname}({', '.join(argtys)})")
+            codegen = utils.gen_codegen(fname)
+            lower(_key, *t.tonumba().args)(codegen)
+
+            return t.tonumba()
+
```
pearu, line 36: This is identical to the `register` function in cmath.py. Provide it from rbc/externals/__init__.py for both cmath.py, libdevice.py, and possibly other externals modules.

pearu: Ah, I see the problem is in `globals`... but then define:
```python
def register(...):
    ...
    return _key
```
and in externals modules use:
```python
for fname, ...:
    globals()[fname] = register(fname, retty, argtypes)
```

rbc/externals/utils.py
```diff
@@ -0,0 +1,25 @@
+from rbc.targetinfo import TargetInfo
```
pearu, line 1: These functions could also live in `rbc/externals/__init__.py` or in `rbc/externals/irtools.py` or here? Your's choice.
Just saying because in the above I suggested moving `register` functions to `__init__.py`.

rbc/irtools.py
```diff
@@ -136,9 +136,12 @@ def set_env(self, env_name, env):
 
 class JITRemoteTypingContext(typing.Context):
     def load_additional_registries(self):
-        from rbc.externals import math
+        from rbc.externals import cmath, libdevice
+        from rbc.omnisci_backend import mathimpl
```
pearu, line 140: These imports have global effect (in the sense that they all exists in `sys.modules`), so I suggest moving these imports out of these method definitions.

In fact, we could do the registration automatically:
```python
import rbc.externals
    ...
    def load_additional_registries(self):
    for module in rbc.externals.values():
        if not isinstance(module, py_types.ModuleType):
            continue
        typing_registry = getattr(module, 'typing_registry', None)
        if typing_registry is not None:
            self.install_registry(typing_registry)
```

In this way, users can control which externals libraries to include by executing:
```
import rbc.externals.<modulename>
```
in their code. Although, this is going to be still have a global effect..

In sum, having the submodule imports in `rbc/externals/__init__.py` will allow introducing new external modules without editing irtools.py. Users can do that by:
```python
sys.modules['rbc.externals.mylib'] = <mylib module object>
```
(we may want to introduce `register_as_external(<module object>)`) as end-user interface but internally use the sys.modules trick.

rbc/targetinfo.py
```diff
@@ -398,7 +398,14 @@ def sizeof(self, t):
             return
         if isinstance(t, str):
             if t == 'int':
+                warnings.warn('Using sizeof(int) == 4')
                 return 4  # this is a guess
+            if t == 'long':
+                warnings.warn('Using sizeof(long) == 4')
+                return 4  # this is a guess
```
pearu, line 405: I suggest using 8:
```python
>>> import ctypes
>>> ctypes.sizeof(ctypes.c_long)
8
```
```

**Review by pearu — CHANGES_REQUESTED** [review]
Thanks, @guilhermeleobas !
Looks good, I had only a couple of structuring suggestions.

File: rbc/targetinfo.py (hunk 5 above)
Comment (guilhermeleobas) [reply rbc/targetinfo.py:405]: I used the following table to derive those numbers: https://en.cppreference.com/w/cpp/language/types#Integer_types

File: rbc/externals/libdevice.py (hunk 2 above)
Comment (guilhermeleobas) [reply rbc/externals/libdevice.py]: Done!

File: rbc/externals/cmath.py [hunk 6]
```diff
@@ -0,0 +1,211 @@
...
+    ),
+}
+
+
+def register(fname, retty, argtys):
+
+    # expose
+    s = f"def {fname}(*args): pass"
+    exec(s, globals())
+    _key = globals()[fname]
```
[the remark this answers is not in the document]
guilhermeleobas: Done! Do you know if there a way for `py_types.FunctionType` include the function in a given global scope? Currently I am doing this way:

```python
# expose
_key = py_types.FunctionType((lambda *args: None).__code__, {}, fname)
_key.__module__ = __name__
globals()[fname] = _key
# ...
module_globals[fname] = _key
_key.__module__ = module_name
del globals()[fname]
```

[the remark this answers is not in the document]
pearu: Why note use:
```
_key = py_types.FunctionType((lambda *args: None).__code__, {}, fname)
module_globals[fname] = _key
```
?
Another approach to get module_globals is to use https://docs.python.org/3/library/sys.html#sys._getframe at one level up.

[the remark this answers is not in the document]
guilhermeleobas: > Why note use:
This end up causing a circular import.

rbc/irtools.py
guilhermeleobas: Done! I've created issue #298 to track `register_as_external` feature.

pearu (approved):
Thanks, @guilhermeleobas , LGTM!

There are a couple of minor nits that we can fix in subsequent PRs. I'll open issues for these.

For the future, I have a couple of notes:
- the functionality provided in rbc.externals should be thoroughly tested using RemoteJIT, not RemoteOmnisci. Although, there can be a couple of tests that would use RemoteOmnisci.
-  for the above, we should create a CI task that runs RemoteJIT server that may be using different numba/python versions and run on different OS (to test Linux/Windows server/client and client/server workflows). Currently we run RemoteJIT server locally and that may give false-positive testing results.
- registering rbc.externals modules must depend on TargetInfo instance properties.

rbc/external.py
```diff
@@ -8,19 +8,26 @@
...
+        except RuntimeError:
+            targetinfo = TargetInfo.dummy()
+
+        # If we already have a targetinfo registered, then, use an empty context manager
+        try:
+            with targetinfo:
+                pass
+        except AssertionError:
+            import contextlib
+            targetinfo = contextlib.nullcontext()
```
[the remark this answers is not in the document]
pearu: Resolving this via #299 .

rbc/external.py
pearu: See #299 .

rbc/externals/cmath.py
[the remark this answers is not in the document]
pearu: Resolving via #300

PR #287, opened 2021-02-19
Add a CHANGELOG file to RBC

guilhermeleobas:
While browsing xarray project, I saw they have a pretty good changelog. Might be a good addition to RBC to have something similar.

There's also a job on github action that reminds us to update the changelog. To skip this, just add the `Skip-Changelog` label to the PR.


pearu:
Would it be possible to have the change log generated from the pull request titles? For categorization, we could use labels.

guilhermeleobas:
> Would it be possible to have the change log generated from the pull request titles? For categorization, we could use labels.

I guess so but I couldn't find an action that works. @goanpeca wants to build one from scratch 😄

pearu:
IIRC, the Quansight github mirroring system was able to pick up details from issues when formatted accordingly. @rgommers ?

pearu:
SymPy uses its bot to generate release notes, see
https://github.com/sympy/sympy-bot
https://github.com/sympy/sympy/wiki/Writing-Release-Notes

rgommers:
> IIRC, the Quansight github mirroring system was able to pick up details from issues when formatted accordingly. @rgommers ?

That doesn't seem too useful here, too heavyweight. Just a general "include all PRs and closed issues in a text file, there's https://github.com/scipy/scipy/blob/master/tools/gh_lists.py.

For more custom things, based on a tag in PR title or something, I believe the SymPy bot is nice indeed.

guilhermeleobas:
Found an action that seems promissing: https://github.com/mikepenz/release-changelog-builder-action

guilhermeleobas:
Change the PR to use an action that auto generate changelog messages using PR titles and labels. This action is triggered when one pushes a new tag to RBC.

You can see an example of such action here: https://github.com/guilhermeleobas/release-changelog-test/releases

I hope everything works as expected in the next RBC release :D

changelog_config.json
```diff
@@ -0,0 +1,62 @@
...
+            "labels": [
+                "documentation"
+            ]
+        },
+        {
+            "title": "## ⚙️ OmniSciDB",
+            "labels": [
+                "omniscidb"
+            ]
+        }
```
pearu, line 32: The omniscidb label is applied to those RBC issues that solution ought to be implemented in omniscidb rather than in rbc. That said, there are omniscidb related PRs in rbc that may implement tests for omniscidb bug fix PRs or that pair with some omniscidb enhancement PR (for instance, when a omniscidb PR introduced the Column feature, there was a PR that implemented Column support in rbc).

However, I think instead of grouping rbc PRs with omniscidb label we should use enhancement or bug fix label for such PRs because it may be hard to decide whether omniscidb label should be applied or not (most changes in rbc are omniscidb related in the current development phase).

So, I suggest removing "OmniSciDB" section.

pearu: Should we have "Testing" section?

pearu (approved):
LGTM

Btw, when tagging and the changelog file is generated, is it possible to regenerate the changelog? Say, when one wants to adjust the labels of PRs?

guilhermeleobas:
> Btw, when tagging and the changelog file is generated, is it possible to regenerate the changelog? Say, when one wants to adjust the labels of PRs?

hum...I don't think so. The changelog is generated based on PRs between tags. One can adjust the labels of a PR at any time before a new tag is pushed to github.

changelog_config.json
guilhermeleobas: I'll replace the OmniSciDB tag by testing, then

guilhermeleobas:
Added another job that checks if the PR has one of the following labels:
- deprecation
- enhancement
- bug
- documentation
- tests
- notebook

pearu (approved):
LGTM. I'll look forward to seeing this in action!

PR #294, opened 2021-02-26
Add JITRemoteCallConv

### LINKED ISSUE #293: Implement a calling convention for remotejit and omnisci targets
Numba has a set of classes that implements the calling convention. Currently, RBC uses the [CPU calling convention](https://github.com/numba/numba/blob/3ad1d914c95883ee4e4fa1175e9afdc0fe9ad31c/numba/core/callconv.py#L326-L340) (default) which does the following transformation

```python
# From
<Python return type> (<Python arguments>)

# To
retcode_t (<Python return type>*, excinfo **, ... <Python arguments>)
```

`excinfo` is a structure that numba uses to pass and retrieve runtime information. Given that we don't need this, we can replace the current calling convention by one of the form:

```python
# From
<Python return type> (<Python arguments>)

# To
retcode_t (<Python return type>*, ... <Python arguments>)
```

Note that we already implement some sort of calling convention on [irtools.py](https://github.com/xnd-project/rbc/blob/6cab3729b13cc9ef71d01c58845afc465c586f32/rbc/irtools.py#L180-L253)

guilhermeleobas:
Fixes #293

### DIFF
--- a/rbc/irtools.py
+++ b/rbc/irtools.py
@@ -11,17 +11,17 @@ from .targetinfo import TargetInfo
 from .utils import get_version
 from .errors import UnsupportedError
 from . import libfuncs
-
 if get_version('numba') >= (0, 49):
     from numba.core import codegen, cpu, compiler_lock, \
         registry, typing, compiler, sigutils, cgutils, \
-        extending
+        extending, utils, callconv
     from numba.core import types as nb_types
     from numba.core import errors as nb_errors
 else:
     from numba.targets import codegen, cpu, registry
     from numba import compiler_lock, typing, compiler, \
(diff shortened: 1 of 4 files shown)


guilhermeleobas:
Notes for the future: It doesn't seems to be possible to remove the `Status` argument from the calling convention without rewriting a lot of internal stuff. i.e. 
* https://github.com/numba/numba/blob/8dd28abf57bd554fe25921ad2abee5d2faa81816/numba/core/imputils.py#L197
* https://github.com/numba/numba/blob/8dd28abf57bd554fe25921ad2abee5d2faa81816/numba/core/base.py#L899

PR #306, opened 2021-03-04
Implement ColumnList support for runtime UDTFs

pearu:
As in title.

Requires https://github.com/omnisci/omniscidb-internal/pull/5274

### DIFF
--- a/rbc/tests/test_omnisci_column_list.py
+++ b/rbc/tests/test_omnisci_column_list.py
@@ -42,8 +42,18 @@ def define(omnisci):
                     out[i] += col[i]
         return lst.size
 
+    @omnisci('int32 columns_sum3(Cursor<Column<int64>, ColumnList<T>>,'
+             ' RowMultiplier, OutputColumn<T>)',
+             T=scalar_types)
+    def columns_sum3(rowid, lst, m, out):
+        for j in range(lst.ncols()):
+            col = lst[j]
+            for i in range(lst.nrows()):
+                out[i] = col[i] if j == 0 else out[i] + col[i]
+        return lst.nrows()
 
-@pytest.mark.parametrize("variant", ['1', "2"])
+
+@pytest.mark.parametrize("variant", ['1', "2", "3"])
 @pytest.mark.parametrize("T", scalar_types)
 def test_columns_sum(omnisci, T, variant):
     v = table_columns_map[T]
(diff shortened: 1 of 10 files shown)


rbc/tests/test_omnisci_column_list.py
```diff
@@ -0,0 +1,63 @@
...
+
+    @omnisci('int32 columns_sum1(Cursor<Column<int64>, ColumnList<T>>,'
+             ' RowMultiplier, OutputColumn<T>)',
+             T=scalar_types)
+    def columns_sum1(rowid, lst, m, out):
+        for i in range(lst.size):
+            out[i] *= 0
+            for j in range(lst.length):
+                out[i] += lst.ptrs[j][i]
+        return lst.size
```
guilhermeleobas, line 29: At least for me, `lst.size` and `lst.length` are a bit confusing. Does the following are equivalent?

```python
len(ColumnList) == ColumnList.length
len(ColumnList[i]) == ColumnList.size
```

rbc/tests/test_omnisci_column_list.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+@pytest.mark.parametrize("variant", ['1', "2"])
+@pytest.mark.parametrize("T", scalar_types)
+def test_columns_sum(omnisci, T, variant):
+    v = table_columns_map[T]
+    c = sql_type_map[T]
+
+    query = f'select rowid, ({v} + {v} + 1) from {omnisci.table_name} ORDER BY rowid'
```
guilhermeleobas, line 52: Should this query contain a `ORDER BY rowid LIMIT 1000` as well?

guilhermeleobas (approved):
Hi Pearu, code looks good to me. Just some minor comments

rbc/tests/test_omnisci_column_list.py
pearu: Not really as the number of rows in the test tables is only 5 and using `LIMIT 1000` is required only to workaround omniscidb bug (without it the server would crash, the reason is unknown atm).

rbc/tests/test_omnisci_column_list.py
guilhermeleobas: Sounds good!

guilhermeleobas:
Hi Pearu, can you write some docs as well explaining how to use and internal attributes about ColumnList? You can put those under `doc/` folder. Don't worry if the docs don't build as expected, I can fix them in another PR.

rbc/tests/test_omnisci_column_list.py
pearu: Resolving via introduction of ColumnList methods `nrows()` and `ncols()`.

PR #315, opened 2021-03-31
Implement tests for UDTF compositions

pearu:
Requires https://github.com/omnisci/omniscidb-internal/pull/5440

### DIFF
--- a/rbc/tests/test_omnisci_udtf.py
+++ b/rbc/tests/test_omnisci_udtf.py
@@ -14,54 +14,56 @@ def define(omnisci):
 
     T = ['int64', 'int32', 'int16', 'int8', 'float64', 'float32']
 
-    @omnisci('int32(Cursor<int64, T>, T, RowMultiplier,'
-             ' OutputColumn<int64>, OutputColumn<T>)', T=T)
-    def sqlmultiply(rowid, col, alpha, row_multiplier, rowid_out, out):
-        for i in range(len(col)):
-            j = rowid[i]
-            out[j] = col[i] * alpha
-            rowid_out[j] = i
-        return len(col)
+    if omnisci.version >= (5, 6):
 
-    @omnisci('int32(Cursor<int64, T>, T, RowMultiplier,'
-             ' OutputColumn<int64>, OutputColumn<T>)', T=T)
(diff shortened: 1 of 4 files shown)


rbc/tests/test_omnisci_udtf.py
```diff
@@ -1,49 +1,140 @@
...
-        'CREATE TABLE IF NOT EXISTS {table_name} ({table_defn});'
-        .format(**locals()))
+    for o in omnisci_fixture(globals()):
+        define(o)
+        yield o
+
+
+def define(omnisci):
+
+    T = ['int64', 'int32', 'int16', 'int8', 'float64', 'float32']
```
guilhermeleobas, line 15: I suppose you don't really need to test all combinations of types given you're testing if function combination works in omniscidb.

rbc/tests/test_omnisci_udtf.py
```diff
@@ -1,49 +1,140 @@
...
+                if j1 == j2:
+                    out[j1] = col1[i1] * col2[i2]
+                    rowid_out[j1] = i1
+                    break
+        return len(col1)
+
+
+@pytest.mark.parametrize("kind", ['i8', 'i4', 'i2', 'i1', 'f8', 'f4'])
+def test_composition(omnisci, kind):
+    omnisci.require_version((5, 6), 'Requires omniscidb-internal PR 5440')
```
guilhermeleobas, line 64:
```suggestion
    omnisci.require_version((5, 7), 'Requires omniscidb-internal PR 5440')
```
The omniscidb-internal version is 5.7, but this always confuses me.

guilhermeleobas (approved):
Thanks @pearu. Code looks good to me, just a couple of comments.

rbc/tests/test_omnisci_udtf.py
pearu: true, but one never knows what will break :) Skipping some types will give only a minor decrease in testing time.

rbc/tests/test_omnisci_udtf.py
pearu: yeah, it seems the omniscidb-internal master is always ahead by 0.1 of an upcoming release version.

PR #330, opened 2021-05-15
Fix test_omnisci_array_null tests

pearu:
As in title.

Just wondering why this test did not fail before.


pearu:
https://github.com/omnisci/omniscidb-internal/pull/5465 fixed the omniscidb bug.
The reason for non-failure is that the tests were wrong.

guilhermeleobas (approved):
LGTM. Tests are passing using on omniscidb-internal master

PR #333, opened 2021-05-16
Address reviewer comments for PR 313

guilhermeleobas:


rbc/tests/test_omnisci_column_default.py
```diff
@@ -83,7 +83,7 @@ def test_default_sizer(omnisci, suffix, kind, expected):
     query = f"select * from table({fn}("
     for i, n in enumerate(kind):
         cols = codes[n]
-        if str.isdigit(n):
+        if n.isdigit:
```
pearu, line 86: `isdigit` is a method, so use `n.isdigit()`.

I wonder if the tests fail when using the master branch of omniscidb-internal?

pearu (changes requested):
Fix a bug.

pearu:
https://github.com/xnd-project/rbc/pull/330 will fix the failing test.

guilhermeleobas:
Ok, I'll wait until #330 gets merged before merging this PR

pearu (approved):
LGTM

pearu:
Requires a rebase.

While it would be fine to merge immediately after rebase, let's wait that all tests pass to ensure that our testing framework functions correctly.

.conda/environment.yml
```diff
@@ -17,7 +17,7 @@ dependencies:
   - tblib
   - thriftpy2
   - netifaces
-  - clang
-  - clangxx
+  - clang=11.1
+  - clangxx=11.1
```
pearu, line 21: This is a suboptimal solution: sooner or later we have to change this again.

guilhermeleobas: I've created https://github.com/xnd-project/rbc/issues/334 to track this. Also, LLVM 11 works by lucky. Officially, llvmlite supports LLVM 10 on most architectures and LLVM 9 on arm64.

pearu: Ok. Since both numba and clang are used in client-side, we should ensure that in conda environment that both use the same llvm version. I don't know if this can be achieved at environment.yml level. At least, we can acquire the clang version from rbc and then compare that with the llvm version that llvmlite uses.

guilhermeleobas: The environment.yml will always use the latest version of Numba. We can stick a LLVM version for clang(xx) that is compatible with Numba. What do you think?

pearu: Sounds good. It likely means that we'll need to stick to some older version of clang that uses llvm that a numba version that rbc minimally supports.

PR #341, opened 2021-06-08
Support output column allocation within a table function

guilhermeleobas:
Requires https://github.com/omnisci/omniscidb-internal/pull/5561 and https://github.com/omnisci/omniscidb-internal/pull/5765 (was: https://github.com/omnisci/omniscidb-internal/pull/5640)

* Pass columns by reference if omniscidb server >= (5, 7)
* Add support for the new sizer `kTableFunctionSpecifiedParameter`
* Implements `set_output_row_size`

### DIFF
--- a/.github/workflows/rbc_test.yml
+++ b/.github/workflows/rbc_test.yml
@@ -210,7 +210,7 @@ jobs:
         run: |
           mamba run -n rbc pytest -sv -r A rbc/tests/ -x -k omnisci
 
-      - name: Run rbc tests [conda]
+      - name: Run rbc tests
         shell: bash -l {0}
         if: ${{ ! ( matrix.os == 'ubuntu-latest' && matrix.omniscidb-from == 'docker' ) }}
         env:
@@ -247,11 +247,3 @@ jobs:
         run: |
           mamba run -n docker docker-compose logs --no-color --tail=10000 -f -t \> omniscidb-docker.log
           cat omniscidb-docker.log
-
-      - name: Stop Omniscidb server [docker]
-        shell: bash -l {0}
-        timeout-minutes: 1
-        if: ${{ matrix.os == 'ubuntu-latest' && matrix.omniscidb-from == 'docker' }}
-        run: |
-          mamba run -n docker docker-compose stop
-          sleep 10
--- a/.github/workflows/rbc_test.yml
+++ b/.github/workflows/rbc_test.yml
@@ -9,12 +9,6 @@ on:
   pull_request:
     branches:
       - master
-  # workflow_dispatch:
-  #   inputs:
-  #     debug_enabled:
-  #       description: 'Run the build with tmate debugging enabled (https://github.com/marketplace/actions/debugging-with-tmate)'
-  #       required: false
-  #       default: false
 
 jobs:
   lint:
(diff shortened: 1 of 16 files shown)


rbc/tests/test_omnisci_array_null.py
```diff
@@ -38,7 +38,7 @@ def test_array_null(omnisci, col):
     if col == 'b':
         omnisci.require_version((5, 7, 0),
                                 'Requires omniscidb-internal PR 5492 [rbc issue 245]',
-                                label='master')
+                                label='docker-dev')
```
guilhermeleobas, line 41: it should be `master` here

pearu: Can you fix it?

pearu (changes requested):
Thanks, @guilhermeleobas ! Looks good. I have one question re `get_or_insert_named_metadata`.

In general, I am very wary about using try-except blocks and recommend explicit handling of whatever the try-except is trying to cover.

rbc/irtools.py
```diff
@@ -345,6 +346,23 @@ def compile_instance(func, sig,
     return fname
 
 
+def get_or_insert_named_metadata(module, name):
+    try:
+        return module.get_named_metadata(name)
+    except KeyError:
+        return module.add_named_metadata(name)
```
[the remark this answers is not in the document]
guilhermeleobas: This is not related to Numba 0.54, but how llvmlite handles metadata. See: https://llvmlite.readthedocs.io/en/latest/user-guide/ir/modules.html#llvmlite.ir.Module.get_named_metadata

rbc/omnisci_backend/omnisci_column.py
```diff
@@ -21,7 +21,12 @@
 class OmnisciColumnType(OmnisciBufferType):
     """Omnisci Column type for RBC typesystem.
     """
-    pass_by_value = True
+    @property
+    def pass_by_value(self):
+        omnisci_version = TargetInfo().software[1][:3]
+        if omnisci_version < (5, 7, 0):
+            return True
+        return False
```
pearu, line 29: nit: replace with
```
return omnisci_version < (5, 7, 0)
```

rbc/tests/test_omnisci_array_null.py
pearu: I gather it should be `docker-dev`?

.github/workflows/rbc_test.yml
```diff
@@ -256,8 +190,20 @@ jobs:
...
         run: |
-          cat omniscidb-output.txt
+          mamba run -n docker docker-compose stop
+          sleep 10
+
+      - name: Run RemoteJIT tests
+        shell: bash -l {0}
+        env:
+          EXPECTED_PYTHON_VERSION: ${{ matrix.python-version }}
+          EXPECTED_OMNISCIDB_VERSION: ${{ matrix.omniscidb-version }}
```
pearu, line 205: nit: Running omnisci related tests are disabled here, hence there is no need to define `EXPECTED_OMNISCIDB_VERSION`

.github/workflows/rbc_test.yml
```diff
@@ -256,8 +190,20 @@ jobs:
...
         shell: bash -l {0}
-        if: matrix.os == 'ubuntu-latest' && matrix.omniscidb-from == 'conda'
+        timeout-minutes: 1
+        if: ${{ matrix.os == 'ubuntu-latest' && matrix.omniscidb-from == 'docker' }}
         run: |
-          cat omniscidb-output.txt
+          mamba run -n docker docker-compose stop
+          sleep 10
+
+      - name: Run RemoteJIT tests
```
pearu, line 201: Does this mean that RemoteJIT tests are run 3 times for each omniscidb version?

It seems that this change disables testing RemoteJIT under Windows and MacOS. Can you restore it? If the matrix grows two big, then restore at least some combinations so that we know that rbc is functional under non-Linux OSs as well.

pearu (changes requested):
LGTM re the aim of the PR but it looks like the PR drops support testing rbc under Windows and MacOS.. Can we restore it?

.github/workflows/rbc_test.yml
guilhermeleobas: Yep, just restored it

Issue #245, opened 2021-01-07
OmniSciDB converts boolean NULL to true


guilhermeleobas:
```sql
omnisql> CREATE TABLE IF NOT EXISTS test (b BOOLEAN[]);
omnisql> INSERT INTO test VALUES ({'true', NULL, 'false'});
omnisql> SELECT * FROM test;
b
{true, true, false}
```
It should be `{true, NULL, false}`

guilhermeleobas:
```sql
omnisql> create table test (i2 SMALLINT[]);
omnisql> insert into test values ({});
omnisql> insert into test values ({NULL});
omnisql> insert into test values ({2});
omnisql> insert into test values ({1, NULL, 3});
omnisql> insert into test values (NULL);
omnisql> select * from test;
i2
{}
{NULL}
{2}
{1, NULL, 3}
NULL
```

from RBC:
```python
>>> _, result = omnisci.sql_execute(f'select i2 from test;')
>>> print(list(result))
[([],), ([-32768],), ([2],), ([1, -32768, 3],), (None,)]
```

pearu:
https://github.com/omnisci/omniscidb-internal/pull/5492/commits/379b5a5c90e149610c2fb0c73bc0d20d3d7516a3 fixes the issue.

pearu:
Using the current omniscidb master:
```
omnisql> CREATE TABLE IF NOT EXISTS test (b BOOLEAN[]);
omnisql> INSERT INTO test VALUES ({'true', NULL, 'false'});
omnisql> SELECT * FROM test;
b
{true, NULL, false}
```

PR #344, opened 2021-06-21
Fix incorrect test result - boolean array containing nulls.

pearu:
Requires https://github.com/omnisci/omniscidb-internal/pull/5492 that implements the fix to #245 

UPDATE: requires_version has new API: removed date and hash, introduced level kw arguments.

### DIFF
--- a/.github/workflows/rbc_test.yml
+++ b/.github/workflows/rbc_test.yml
@@ -200,7 +200,7 @@ jobs:
         shell: bash -l {0}
         if: ${{ failure() && matrix.os == 'ubuntu-latest' && matrix.omniscidb-from == 'docker' }}
         run: |
-          test -f omniscidb.pid && ps -F --pid `cat omniscidb.pid`  # to check if server is still alive
+          mamba run -n docker docker-compose ps  # to check if server is still alive
           mamba run -n docker docker-compose logs --no-color -f -t --tail=10000 \> omniscidb-docker.log
           cat omniscidb-docker.log
 
--- a/.github/workflows/rbc_test.yml
+++ b/.github/workflows/rbc_test.yml
@@ -192,7 +192,7 @@ jobs:
         shell: bash -l {0}
         if: matrix.os == 'ubuntu-latest' && matrix.omniscidb-from == 'docker'
         env:
-          OMNISCIDB_DEV_LABEL: 'docker-dev'
+          OMNISCIDB_DEV_LABEL: ${{ 'docker-dev' }}
           EXPECTED_PYTHON_VERSION: ${{ matrix.python-version }}
           EXPECTED_OMNISCIDB_VERSION: ${{ matrix.omniscidb-version }}
         run: |
@@ -209,7 +209,7 @@ jobs:
 
       - name: Show Omniscidb docker logs on failure [docker]
         shell: bash -l {0}
-        if: ${{ failure() && matrix.os == 'ubuntu-latest' && matrix.omniscidb-from == 'docker' }}
+        if: failure() && matrix.os == 'ubuntu-latest' && matrix.omniscidb-from == 'docker'
         run: |
           mamba run -n docker docker-compose top  # to check if server is still alive
           mamba run -n docker docker-compose exec -T ls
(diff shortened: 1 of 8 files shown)


.github/workflows/rbc_test.yml
```diff
@@ -200,8 +201,10 @@ jobs:
         shell: bash -l {0}
         if: ${{ failure() && matrix.os == 'ubuntu-latest' && matrix.omniscidb-from == 'docker' }}
         run: |
-          mamba run -n docker docker-compose logs -f -t \> omniscidb-docker.log
-          cat omniscidb-docker.log
+          mamba run -n docker docker-compose top  # to check if server is still alive
+          mamba run -n docker docker-compose exec -T ls
+          #mamba run -n docker docker-compose logs --no-color -f -t --tail=100 \> omniscidb-docker.log
+          #cat omniscidb-docker.log
```
pearu, line 207: For some reason, this has never been functional.

rbc/tests/test_omnisci_array_null.py
```diff
@@ -34,10 +34,11 @@ def test_array_null(omnisci, col):
     if col in ['i2', 'i8', 'f8']:
         omnisci.require_version((5, 7, 0),
                                 'Requires omniscidb-internal PR 5465 [rbc PR 330]',
-                                hash='4777a06b01')
-    # skipping bool test since NULL is converted to true - rbc issue #245
+                                level=1)
```
guilhermeleobas, line 37: instead of `level=1`, can we name those numbers? For instance:
* `OMNISCIDB_DEV_DOCKER=1`
* `OMNISCIDB_INTERNAL_PR=10`

rbc/tests/test_omnisci_array_null.py
```diff
@@ -34,10 +34,11 @@ def test_array_null(omnisci, col):
...
     if col == 'b':
-        pytest.skip('Skipping Array<boolean> test [RBC issue 240]')
+        omnisci.require_version((5, 7, 0),
+                                'Requires omniscidb-internal PR 5492 [rbc issue 245]',
+                                level=10)
```
guilhermeleobas, line 41: Same here!

rbc/tests/test_omnisci_testing.py
```diff
@@ -12,7 +12,11 @@ def test_table_load(mth, suffix):
     for omnisci in omnisci_fixture(globals(), load_columnar=load_columnar):
         omnisci.require_version((5, 7, 0),
                                 'Requires omniscidb-internal PR 5465 [rbc PR 330]',
-                                hash='4777a06b01')
+                                level=1)
```
guilhermeleobas, line 15: Same

rbc/tests/test_omnisci_testing.py
```diff
@@ -12,7 +12,11 @@ def test_table_load(mth, suffix):
...
+        if load_columnar and suffix == 'arraynull':
+            omnisci.require_version((5, 7, 0),
+                                    'Requires omniscidb-internal PR 5492 [rbc issue 245]',
+                                    level=10)
```
guilhermeleobas, line 19: Same

guilhermeleobas (approved):
PR looks good to me. Just a minor comment

rbc/tests/test_omnisci_array_null.py
pearu: yes, labelling such additional requirements makes sense.

How about:
```
omnisci.require_version(version, message, label='release')
```
where `label` is in `['release', 'dev-docker', 'dev']` (that correspond to current levels 0, 1, and 10, respectively), and the environment may define
```
OMNISCIDB_DEV_LABEL=<label value>
```

pearu: Actually, I am thinking that using the branch name for label makes more sense. When OMNISCIDB_DEV_LABEL and the specified label match, only then the particular test will be run.

guilhermeleobas: That's a better approach

pearu: Fixed. See the doc-string of require_version. In short, when testing rbc against omniscidb master or any branch that is based on master, use
```
export OMNISCIDB_DEV_LABEL=master
```
When implementing a test that requires a specific branch of omniscidb, say, `foo`, then use
```
export OMNISCIDB_DEV_LABEL=foo
```
and in the test, use `omnisci.require_version(..., label='foo')`.

PR #350, opened 2021-07-23
Speedup cmath tests

### LINKED ISSUE #346: test_externals_cmath.py::test_external_cmath_omnisci tests are slow
As in title. The reason is that each subtest compiles its own function to an independent LLVM module. To speed up the tests, all functions should be compiled into a single LLVM module.

guilhermeleobas:
Fixes #346 

Time before
```
===================================== 147 passed, 9 skipped, 9 xfailed, 26 warnings in 295.00s (0:04:54) =====================================

real    4m55.816s
user    0m18.301s
sys     0m7.369s
```

Time after
```
========================================== 147 passed, 9 skipped, 9 xfailed, 26 warnings in 14.62s ===========================================

real    0m15.047s
user    0m7.045s
sys     0m7.224s
```

### DIFF
--- a/rbc/tests/test_externals_cmath.py
+++ b/rbc/tests/test_externals_cmath.py
@@ -107,11 +107,16 @@ cmath_funcs = (
 funcs = {}
 
 
+def _get_query_func(fname):
+    PREFIX = "test_cmath"
+    return f"{PREFIX}_{fname}"
+
+
 def define(jit):
     global funcs
 
-    def inner(fname, signature):
-        cmath_fn = getattr(cmath, fname)
+    def inner(cmath_func, query_func, signature):
+        cmath_fn = getattr(cmath, cmath_func)
         t = Type.fromstring(signature)
         retty = str(t[0])
         argtypes = tuple(map(str, t[1]))
@@ -130,7 +135,7 @@ def define(jit):
             def fn(a, b, c):
                 return cmath_fn(a, b, c)
 
-        fn.__name__ = fname
+        fn.__name__ = query_func
         fn = jit(f"{retty}({', '.join(argtypes)})", devices=["cpu"])(fn)
         return fn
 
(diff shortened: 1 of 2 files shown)


rbc/tests/test_externals_cmath.py
```diff
@@ -122,11 +130,16 @@ def fn(a, b):
             def fn(a, b, c):
                 return cmath_fn(a, b, c)
 
-        fn.__name__ = fn_name
+        fn.__name__ = fname
```
pearu, line 133: We cannot use the same name of an omnisci SQL function for a UDF. For instance, when `fname == 'cos'`, the corresponding UDF will never be called as omnisci already defines `COS` function.
To demonstrate this, apply

```
diff --git a/rbc/tests/test_externals_cmath.py b/rbc/tests/test_externals_cmath.py
index f8a2707..4bf8e20 100644
--- a/rbc/tests/test_externals_cmath.py
+++ b/rbc/tests/test_externals_cmath.py
@@ -106,7 +106,7 @@ cmath_funcs = (
 
 funcs = {}
 
-
+from rbc.externals.stdio import printf
 def define(jit):
     global funcs
 
@@ -116,10 +116,12 @@ def define(jit):
         retty = str(t[0])
         argtypes = tuple(map(str, t[1]))
         arity = len(argtypes)
-
+        bfname = ('CALLING ' + fname + '\n').encode()
         # define callable
         if arity == 1:
             def fn(a):
+                for c in bfname:
+                    printf("%c", c)
                 return cmath_fn(a)
 
         elif arity == 2:
```
and notice that omnisci server output does not contain `CALLING cos`, for instance.

So, we'll need to restore `fn_name` feature to be sure that rbc test suite will not define UDFs with the same names as existing SQL operators.

pearu (changes requested):
This is a great improvement, thanks @guilhermeleobas !

However, we must make sure that we call the defined UDFs, not the existing omnisci SQL operators, such as `sin`, `cos`, etc.

pearu (approved):
LGTM!

Issue #327, opened 2021-04-29
test_columns_sum[float64-3] crashes CUDA enabled omniscidb server 5.7


pearu:
As in title. The test passes when using CUDA disabled server.

Crash message:
```
2021-04-29T15:48:03.720368 E 15779 49 CudaMgr.cpp:169 CUDA Error: CUDA Error (700): an illegal memory access was encountered
2021-04-29T15:48:03.720453 E 15779 49 CudaMgr.cpp:169 CUDA Error: CUDA Error (700): an illegal memory access was encountered
2021-04-29T15:48:03.721019 E 15779 2056 DBHandler.cpp:1414 Exception: CUDA Error (700): an illegal memory access was encountered
```

pearu:
The test `pytest -sv rbc/tests/ -k test_columns_sum` works fine with:
```
OmnisciDB version (5, 8, 0, 'dev-20210907-b3947da3f5')
```
but crashes with
```
OmnisciDB version (5, 7, 0, '-20210907-')
```
Both servers are CUDA enabled.

gdb traceback:
<details>

```
(gdb) bt
#0  0x00007fffe2cae0b0 in columns_sum1.cpu_0 ()
#1  0x00007fffe2cae08f in call_table_function ()
#2  0x00005555561081c1 in TableFunctionExecutionContext::launchCpuCode(TableFunctionExecutionUnit const&, TableFunctionCompilationContext const*, std::vector<signed char const*, std::allocator<signed char const*> >&, std::vector<long, std::allocator<long> >&, unsigned long, Executor*) [clone .localalias] ()
#3  0x00005555561096ab in TableFunctionExecutionContext::execute(TableFunctionExecutionUnit const&, std::vector<InputTableInfo, std::allocator<InputTableInfo> > const&, TableFunctionCompilationContext const*, ColumnFetcher const&, ExecutorDeviceType, Executor*) ()
#4  0x0000555555e73305 in Executor::executeTableFunction(TableFunctionExecutionUnit, std::vector<InputTableInfo, std::allocator<InputTableInfo> > const&, CompilationOptions const&, ExecutionOptions const&, Catalog_Namespace::Catalog const&) ()
#5  0x0000555556023ee1 in RelAlgExecutor::executeTableFunction(RelTableFunction const*, CompilationOptions const&, ExecutionOptions const&, long) [clone .localalias] ()
[...]

PR #361, opened 2021-09-09
Update omniscidb versions in github actions, dropping 5.4 and 5.2.

pearu:
As in the title.

### DIFF
--- a/.github/workflows/rbc_test.yml
+++ b/.github/workflows/rbc_test.yml
@@ -96,7 +96,7 @@ jobs:
         os: [ubuntu-latest]
         python-version: [3.9, 3.8, 3.7]
         numba-version: [0.54, 0.53]
-        omniscidb-version: [5.5, 5.4, 5.2]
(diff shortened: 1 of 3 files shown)


pearu:
https://github.com/omnisci/omniscidb-internal/tree/rc/v5.7.1 contains now https://github.com/omnisci/omniscidb-internal/pull/5765 that will fix https://github.com/xnd-project/rbc/issues/327

.github/workflows/rbc_test.yml
```diff
@@ -113,8 +113,8 @@ jobs:
             omniscidb-from: docker
           - os: ubuntu-latest
             python-version: 3.8
-            omniscidb-version: 5.6.4
-            docker-image: omnisci/core-os-cpu:v5.6.4
+            omniscidb-version: 5.7.0
+            docker-image: omnisci/core-os-cpu:v5.7.0
```
guilhermeleobas, line 117: If we can test omniscidb 5.7 using conda-forge, we don't need to test it again using docker.

Issue #359, opened 2021-08-24
`get_version` endpoint is never cached


guilhermeleobas:
This is a small bug in RBC I just found yesterday. 

https://github.com/xnd-project/rbc/blob/cacf447af6b599c914964b3f2d27fbce430f4b81/rbc/omniscidb.py#L280-L285

`self._version` is never assigned, and the thrift endpoint is always called.

guilhermeleobas:
Fixed in #362

PR #362, opened 2021-09-16
Test parallel execution of table functions

pearu:
- implement tests for https://github.com/omnisci/omniscidb-internal/pull/5901 [QE-99]
- ~introduce register kw argument to sql_execute to avoid unnecessary locking of omniscidb server.~
- implement a smarter way to skip register for queries that don't use pending UDF/UDTFs
- fix rbc issue #359.

### DIFF
--- a/rbc/omniscidb.py
+++ b/rbc/omniscidb.py
@@ -82,14 +82,26 @@ def get_literal_return(func, verbose=False):
                 print(source)
 
 
+def _global_omnisci():
+    """Implements singleton of a global RemoteOmnisci instance.
+    """
+    config = get_client_config()
+    omnisci = RemoteOmnisci(**config)
+    while True:
+        yield omnisci
+
+
+global_omnisci_singleton = _global_omnisci()  # generator object
+
+
 def is_available(_cache={}):
     """Return version tuple and None if OmnisciDB server is accessible or
     recent enough. Otherwise return None and the reason about
     unavailability.
     """
+
     if not _cache:
-        config = get_client_config()
-        omnisci = RemoteOmnisci(**config)
+        omnisci = next(global_omnisci_singleton)
         try:
             version = omnisci.version
         except Exception as msg:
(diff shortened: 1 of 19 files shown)


guilhermeleobas:
Rather than implementing an internal `sql_execute` version with `register=False`, does it make sense to have a version of `sql_execute`, say `sql_execute_no_register` on `rbc/tests/__init__.py`?

pearu:
> Rather than implementing an internal `sql_execute` version with `register=False` on every test file, does it make sense to have a version of `sql_execute`, say `sql_execute_no_register/sql_execute_unlock` on `rbc/tests/__init__.py`?

Such a function would need to use a global `RemoteOmnisci` instance (equivalent to pyomniscidb or pymapd) that may be independent of other `RemoteOmnisci` instances created in various tests files, as well as globally accessible for all test files.
So, yes, that would make sense, although, I'd prefer the shorter name `sql_execute`: it would be a function, and tests would use `sql_execute` method when registration is required.

In the longer run, I think parallel execution of rbc tests would be nice to have in order to maximally stress omniscidb registration end-point and to test UDF/UDTF multi-client deployment.

pearu:
That said, I'll implement global `sql_execute` and remove the `register=False` usages. The changeset will be smaller then as well.

pearu:
@guilhermeleobas I have eliminated `register=` kw from `sql_execute` method as unneeded.

guilhermeleobas:
I mean, implement a method of `RemoteOmnisci` called `sql_execute_unlock`. And to use, one can just call `omnisci.sql_execute_unlock`.

pearu:
> I mean, implement a method of `RemoteOmnisci` called `sql_execute_unlock`. And to use, one can just call `omnisci.sql_execute_unlock`.

Introducing such a method seems unnecessary with the current sql_execute implementation: the register method is called only if the query uses function names that are pending for registration.

guilhermeleobas:
> In the longer run, I think parallel execution of rbc tests would be nice to have in order to maximally stress omniscidb registration end-point and to test UDF/UDTF multi-client deployment.

By using the singleton instance of `RemoteOmnisci`, can we run rbc tests in parallel?

pearu:
> By using the singleton instance of `RemoteOmnisci`, can we run rbc tests in parallel?

Yes! In fact, https://github.com/xnd-project/rbc/pull/362/files#diff-b981a1cbaf736b0eb3e05d718756deeb70be993d465efd38312040bef0d4f57bR226 does exactly this.

guilhermeleobas (approved):

PR #367, opened 2021-10-21
Add support for table_function_error

antocuni:
What the tittle says.

I **think** that `rbc` does not fully support string manipulations, because as soon as we use strings we quickly end up calling `PyUnicode_*` functions. To overcome the limitation, we require the message passed to `table_function_error` to be a string literal, similar to how numba implements `printf`.

### DIFF
--- a/rbc/externals/omniscidb.py
+++ b/rbc/externals/omniscidb.py
@@ -7,6 +7,7 @@ from rbc import irutils
 from rbc.errors import UnsupportedError
 from rbc.targetinfo import TargetInfo
 from numba.core import extending, types as nb_types
+from numba.core.cgutils import make_bytearray, global_constant
 from llvmlite import ir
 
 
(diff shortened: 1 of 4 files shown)


pearu (approved):
LGTM! Thanks @antocuni !

rbc/externals/omniscidb.py
```diff
@@ -36,6 +37,34 @@ def codegen(context, builder, sig, args):
...
 
+@extending.intrinsic
+def table_function_error(typingctx, message):
+    """
+    Return an error from a UDTF.
+
+    ``message`` must be a string literal.
+    """
+    if not isinstance(message, nb_types.StringLiteral):
+        raise TypeError(f"expected StringLiteral but got {type(message).__name__}")
```
guilhermeleobas, line 48: Note for the future: Replace  `TypeError` by `NumbaTypeError` after Numba 0.55 gets released.

guilhermeleobas (approved):
LGTM. Thanks, Antonio

rbc/externals/omniscidb.py
pearu: @guilhermeleobas let's collect these bits into https://github.com/xnd-project/rbc/pull/369 . Breno PR has a similar thing that might land earlier..

PR #369, opened 2021-10-23
Fixes for numba 0.55

### LINKED ISSUE #364: Test the new error-handling model implemented in Numba master
Numba updated its error-handling model in https://github.com/numba/numba/pull/7397. Previously, any error (i.e. `IndexError`) raised by a code mistake was treated as an internal compilation failure and were swallowed by the compiler. So, you couldn't differentiate a bug in the source code from a non function matching. The new model treats this better. 

One should test RBC against Numba main branch to make sure it doesn't break anything when Numba 0.56 gets released. There are instructions in the first post on how to enable the new error handling.

guilhermeleobas:
Fixes #364 

How to test it: Clone the Numba repo and run pytest with `PYTHONPATH=/path/to/numba NUMBA_CAPTURED_ERRORS="new-style" pytest -rs -sv --disable-warnings --tb=line rbc/tests -x`

Changes:
* `external` is implemented as an `intrinsic` which is inserted into the global scope with `exec`
* ~With `@intrinsic`, it is not possible to postpone the _translation_ of some types (i.e. `int`) to after the target info is available. Therefore, some `cmath` signatures now uses client specific size (i.e. `numba.long_`) instead of the server ones. This might be a problem for us~
* No more specific typing/lowering registry for each implementation files. Everything now uses the same built-in registry available from Numba - `numba.core.(imputils/typing.templates).builtin_registry`
* Back port a couple exceptions from Numba 0.55


rbc/externals/__init__.py
```diff
@@ -29,37 +27,28 @@ def sanitize(name):
...
+        f'        return builder.call(func, args)\n'
+        f'    return signature, codegen'
+    )
+
+    exec(fn_str)
+    fn = locals()[fname]
+    fn.__module__ = module_name
+    fn.__doc__ = doc
+    fn.__name__ = fname
+    module_globals[fname] = fn
```
pearu, line 53: Nit: I wonder if this could be rewritten as:
```suggestion
    exec(fn_str, module_globals)
    fn = module_globals[fname]
    fn.__doc__ = doc
```
If yes, `module_name` can be eliminated, otherwise, never mind.

rbc/externals/__init__.py
```diff
@@ -29,37 +27,28 @@ def sanitize(name):
...
+        f'    retty_ = Type.fromobject("{retty}").tonumba()\n'
+        f'    argnames_ = tuple(map(lambda x: Type.fromobject(x).tonumba(), [{", ".join(argnames)}]))\n'  # noqa: E501
+        f'    signature = retty_(*argnames_)\n'
+        f'    def codegen(context, builder, sig, args):\n'
+        f'        from numba.core import funcdesc\n'
+        f'        fndesc = funcdesc.ExternalFunctionDescriptor("{fname}", sig.return_type, sig.args)\n'  # noqa: E501
+        f'        func = context.declare_external_function(builder.module, fndesc)\n'
+        f'        return builder.call(func, args)\n'
+        f'    return signature, codegen'
+    )
```
pearu, line 46: Nit: suggest using multiline string:
```
fn_str = f'''\
from numba.core.extending import intrinsic
@intrinsic
def {fname}(typingctx, {", ".join(argnames)}):
    ...
'''
```

rbc/structure_type.py
```diff
@@ -9,8 +9,8 @@
 
 """ TODO: use local registries, currently blocked by overloading
```
pearu, line 10: Is this TODO item still relevant?

pearu (approved):
I have a couple of nits but otherwise LGTM! Thanks, @guilhermeleobas !

rbc/structure_type.py
guilhermeleobas: No, I'll remove it.

PR #382, opened 2021-11-23
Throw error if raise statements are used

guilhermeleobas:
This PR changes Numba IR and replace `raise exception(msg)` nodes by `return table_function_error(msg)`.

PS: The exception class being raised is not important.

### DIFF
--- a/rbc/omnisci_backend/omnisci_pipeline.py
+++ b/rbc/omnisci_backend/omnisci_pipeline.py
@@ -62,7 +62,6 @@ class CheckRaiseStmts(FunctionPass):
         FunctionPass.__init__(self)
 
     def run_pass(self, state):
-        mutated = False
         func_ir = state.func_ir
         for blk in func_ir.blocks.values():
             for _raise in blk.find_insts(ir.Raise):
@@ -71,7 +70,7 @@ class CheckRaiseStmts(FunctionPass):
                        'to raise an error.')
                 loc = _raise.loc
                 raise NumbaTypeError(msg, loc=loc)
-        return mutated
+        return False
 
 
 class OmnisciCompilerPipeline(CompilerBase):
--- a/rbc/omnisci_backend/omnisci_pipeline.py
+++ b/rbc/omnisci_backend/omnisci_pipeline.py
@@ -1,4 +1,4 @@
-from rbc.externals.omniscidb import table_function_error
+from rbc.errors import NumbaTypeError
 from .omnisci_buffer import BufferMeta, free_omnisci_buffer
 from numba.core import ir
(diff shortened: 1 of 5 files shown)


brenocfg:
Since we always use `table_function_error` calls, I think this implicitly causes all exception-raising Python UDTFs to run sequentially, right? Maybe we could check if the UDTF receives a `TableFunctionManager` as input, and if so call `error_message` on that rather than `table_function_error`?

rbc/omnisci_backend/omnisci_pipeline.py
```diff
@@ -54,6 +54,41 @@ def run_pass(self, state):
...
+                blk.insert_after(fn, glob)
+                call.value.func = fn.target
+
+                # create a return inst
+                ret = ir.Return(value=call.target, loc=loc)
+                blk.append(ret)
+
+                # remove raise instruction
+                blk.remove(_raise)
+
```
antocuni, line 86: At first I was confused by this, because we remove the `_raise` in the middle of the block and we append the `return` at the end.
Then I realized that this works only when the `raise` is *always* the last instruction of the block. I think this is a property that we can assume as guaranteed, but it's better to add an `assert` just in case, before doing the `append/remove` dance:
```python
assert blk.body[-1] is _raise
```

antocuni (commented):
LGTM

rbc/omnisci_backend/omnisci_pipeline.py
```diff
@@ -54,6 +54,26 @@ def run_pass(self, state):
...
 
+@register_pass(mutates_CFG=False, analysis_only=False)
+class CheckRaiseStmts(FunctionPass):
+    _name = "check_raise_stmts"
+
+    def __init__(self):
+        FunctionPass.__init__(self)
+
+    def run_pass(self, state):
+        mutated = False
```
pearu, line 65: Remove as unused
```suggestion
```

rbc/omnisci_backend/omnisci_pipeline.py
```diff
@@ -54,6 +54,26 @@ def run_pass(self, state):
...
+        func_ir = state.func_ir
+        for blk in func_ir.blocks.values():
+            for _raise in blk.find_insts(ir.Raise):
+                msg = ('raise statement is not supported in '
+                       'UDF/UDTFs. Please, use `return table_function_error(msg)` '
+                       'to raise an error.')
+                loc = _raise.loc
+                raise NumbaTypeError(msg, loc=loc)
+        return mutated
```
pearu, line 74:
```suggestion
        return False
```

pearu (approved):
I have one nit but otherwise LGTM! Thanks, @guilhermeleobas!

PR #379, opened 2021-11-12
Introduce numba target extension to RBC

guilhermeleobas:
Numba 0.54 introduced a new API for register new synthetic targets. For RBC, this means that one can implement intrinsics and overload functions/methods that is available to a specific target (i.e. `omniscidb_cpu` or `omniscidb_gpu`). It also solves a long issue (#343) where Numba would use the CPU bindings in the GPU target.

### Summary of changes

* Refactor the target and typing contexts and introduce `OmnisciTargetOptions`, `OmnisciTargetDescr` and specific dispatchers.
* Remove the context manager used to replace the internal codegen by the one OmniSciDB uses. With the introduction of specific dispatchers, this is not needed anymore.

### Things to discuss
* This is a breaking change and it requires Numba 0.54 or greater to work
~* libdevice support seems to be broken in OmniSciDB 5.9 (or even master)~

### To-do
* ~Overloads don't work yet~

PR #383, opened 2021-11-25
Use libdevice fn if available

guilhermeleobas:
Update some bindings to use libdevice function if available


pearu:
Does there exist a test that would fail before this PR? E.g. use math functions in a GPU-only UDF so that when omniscidb wants to switch to CPU execution, it would fail.

guilhermeleobas:
I've moved the code here + tests to #379

PR #385, opened 2021-12-01
Add omniscidb v5.9 to our CI

guilhermeleobas:
As title.

### DIFF
--- a/.github/workflows/rbc_test.yml
+++ b/.github/workflows/rbc_test.yml
@@ -113,7 +113,7 @@ jobs:
             omniscidb-from: docker
           - os: ubuntu-latest
             python-version: 3.8
-            omniscidb-version: 5.8.0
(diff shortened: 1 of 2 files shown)


guilhermeleobas:
@pearu, should we deprecate and remove any code that was written for omniscidb 5.5 or older? i.e. old udtf style, tests are skipped if omniscidb version < 5.6

pearu:
> @pearu, should we deprecate and remove any code that was written for omniscidb 5.5 or older? i.e. old udtf style, tests are skipped if omniscidb version < 5.6

Yes, let's continue using the policy of supporting the last three omniscidb versions that are available from conda-forge that we test rbc against. So, currently, the versions list is 5.6, 5.7, 5.8.

pearu (approved):
LGTM!

PR #386, opened 2021-12-01
Drop support for omniscidb v5.(2/3/4/5)

guilhermeleobas:
As title


rbc/omniscidb.py
```diff
@@ -419,62 +419,6 @@ def load_table_columnar(self, table_name, **columnar_data):
 
         thrift = self.thrift_client.thrift
         table_details = self.get_table_details(table_name)
-        use_sql = self.version[:2] < (5, 3)
```
pearu, line 422: When a given rbc version connects to some older omniscidb version, something will break below. In order to notify that the breakage is not a bug, let's have here at least the following assert statement:
```python
        assert self.version[:2] >= (5, 6), self.version  # omniscidb is too old
```

The above note applies to all changes in this PR that may cause rbc failure when connecting to an unsupported omniscidb version.

pearu (changes requested):
See my comment below. We should ensure that rbc failure from connection to unsupported omniscidb server version is user-friendly and informative: the failure is not due to a bug but unsupported omniscidb version.

guilhermeleobas:
Pearu, I've replaced the exception by a pending deprecation warning. This way, we will not break any existing code using older versions of omniscidb

rbc/omniscidb.py
```diff
@@ -295,6 +295,11 @@ def version(self):
         if self._version is None:
             version = self.thrift_call('get_version')
             self._version = parse_version(version)
+            if self._version[:2] < (5, 6):
+                ver = ".".join(map(str, self._version))
+                msg = (f'OmniSciDB server is too old (v.{ver}) and some '
```
pearu, line 300: Nit:
```suggestion
                msg = (f'OmniSciDB server v.{version} is too old (expected v.5.6 or newer) and some '
```

pearu (approved):
I have a minor nit but LGTM! Thanks, @guilhermeleobas!

PR #387, opened 2021-12-07
Remove `label=` from some tests

guilhermeleobas:
This PR removes the `label=...` of a few tests that were not being executed in a pytest run. I've tested it on 
* omniscidb 5.7 cpu/gpu
* omniscidb 5.8 cpu/gpu
* omniscidb 5.9 cpu/gpu
* omniscidb master cpu/gpu

### DIFF
--- a/rbc/tests/test_omnisci_udtf.py
+++ b/rbc/tests/test_omnisci_udtf.py
@@ -131,7 +131,7 @@ def test_composition(omnisci, kind):
 
 
 def test_table_function_manager(omnisci):
-    omnisci.require_version((5, 9), 'Requires omniscidb-internal PR 6035', label='master')
+    omnisci.require_version((5, 9), 'Requires omniscidb-internal PR 6035')
 
     @omnisci('int32(TableFunctionManager, Column<double>, OutputColumn<double>)')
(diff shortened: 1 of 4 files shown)


pearu (approved):
LGTM!

rbc/tests/test_omnisci_udtf.py
```diff
@@ -169,7 +169,8 @@ def test_parallel_execution(omnisci, sleep, mode):
     (ct_sleep1).
 
     """
-    omnisci.require_version((5, 8), 'Requires omniscidb-internal PR 5901')
+    omnisci.require_version((5, 8), 'Requires omniscidb-internal PR 5901',
```
pearu: https://github.com/omnisci/omniscidb-internal/pull/5901 has been merged. I don't know if the corresponding changeset is in 5.8 or 5.9, but if it is not, we should not use the label as the changeset is in master.

guilhermeleobas: Just tested and `ct_sleep` is not on omniscidb 5.8 but it is on 5.9. I'll update the PR accordingly.

pearu (approved):
LGTM!

PR #388, opened 2021-12-07
Change the return value if the array is empty for `max`

### LINKED ISSUE #114: test_ndarray_methods[max_empty] does not work with CPU-only omniscidb v 5.4
`test_ndarray_methods[max_empty]` fails (return value of the UDF is `None`, expecting -128) while the omniscidb server reports at the exit:
```
omnisci_array_max: cannot find max of zero-sized array
```
OmniSci Version: 5.4.0dev-20200701-9b87e0f99b

guilhermeleobas:
The [array API](https://data-apis.org/array-api/latest/API_specification/statistical_functions.html#max-x-axis-none-keepdims-false) says the maximum value when the container is empty is implementation specific. In RBC case, we choose to return the minimum value for the container type, but this collides with the value omniscidb uses to signal a null value.

For reference, below is the set of values omniscidb uses as a sentinel for null values.
```javascript
{
    'boolean8': -128,
    'int8': -128,
    'int16': -32768,
    'int32': -2147483648,
    'int64': -9223372036854775808,
    'uint8': 255,
    'uint16': 65535,
    'uint32': 4294967295,
    'uint64': 18446744073709551615,
    'float32': 1.1754944e-38,
    'float64': 2.2250738585072014e-308,
    'Array<boolean8>': -127,
    'Array<int8>': -127,
    'Array<int16>': -32767,
    'Array<int32>': -2147483647,
    'Array<int64>': -9223372036854775807,
    'Array<float32>': 2.3509887e-38,
    'Array<float64>': 4.450147717014403e-308
}
```

I'm open for suggestions for other values to use as sentinel

Fixes #114

### DIFF
--- a/rbc/omnisci_backend/numpy_funcs.py
+++ b/rbc/omnisci_backend/numpy_funcs.py
@@ -195,13 +195,20 @@ def omnisci_array_fill(x, v):
 @extending.overload_method(ArrayPointer, 'max')
 def omnisci_array_max(x, initial=None):
     if isinstance(x, ArrayPointer):
-        min_value = get_type_limits(x.eltype).min
+        # the array api standard says this is implementation specific
+        limits = get_type_limits(x.eltype)
+        t = typesystem.Type.fromobject(x.eltype)
+        if t.is_float:
+            min_value = limits.min
+        elif t.is_int or t.is_uint:
+            min_value = 0 if t.is_uint else limits.min + 1
+        else:
+            raise TypeError(f'Unsupported type {t}')
 
         def impl(x, initial=None):
             if len(x) <= 0:
                 printf("omnisci_array_max: cannot find max of zero-sized array")  # noqa: E501
-                # the array api standard says this is implementation specific
-                return min_value + 1
+                return min_value
             if initial is not None:
                 m = initial
             else:
@@ -272,10 +279,7 @@ def omnisci_np_prod(a, initial=None):
 @expose_and_overload(np.mean)
 @extending.overload_method(ArrayPointer, 'mean')
 def omnisci_array_mean(x):
-    if isinstance(x.eltype, types.Integer):
-        zero_value = 0
-    elif isinstance(x.eltype, types.Float):
-        zero_value = np.nan
+    zero_value = np.nan  # 0 if x.dtype is int
 
     if isinstance(x, ArrayPointer):
         def impl(x):
--- a/rbc/omnisci_backend/numpy_funcs.py
+++ b/rbc/omnisci_backend/numpy_funcs.py
@@ -279,7 +279,7 @@ def omnisci_np_prod(a, initial=None):
 @expose_and_overload(np.mean)
 @extending.overload_method(ArrayPointer, 'mean')
 def omnisci_array_mean(x):
-    zero_value = np.nan  # 0 if x.dtype is int
+    zero_value = np.nan
 
     if isinstance(x, ArrayPointer):
         def impl(x):
--- a/rbc/omnisci_backend/numpy_funcs.py
+++ b/rbc/omnisci_backend/numpy_funcs.py
@@ -200,7 +200,8 @@ def omnisci_array_max(x, initial=None):
         def impl(x, initial=None):
             if len(x) <= 0:
                 printf("omnisci_array_max: cannot find max of zero-sized array")  # noqa: E501
-                return min_value
+                # the array api standard says this is implementation specific
+                return min_value + 1
             if initial is not None:
                 m = initial
             else:
--- a/rbc/tests/test_omnisci_array_methods.py
+++ b/rbc/tests/test_omnisci_array_methods.py
@@ -9,6 +9,9 @@ available_version, reason = rbc_omnisci.is_available()
 pytestmark = pytest.mark.skipif(not available_version, reason=reason)
 
 
+NUMERIC_TYPES = ['int8', 'int16', 'int32', 'int64', 'float32', 'float64']
+
+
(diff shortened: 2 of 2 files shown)


pearu:
For signed integers, I suggest using the `minimal value + 1` for `max` reduction.
For unsigned integers, return 0.
For floats, return `numpy.finfo(dtype).min`.

Keep in mind that for a reduction operation op, the reduction identity i is such that `op([i] + args) == op(args)` where `args` is a sequence of value over which reduction is computed. `op([])` should return `i`.

rbc/omnisci_backend/numpy_funcs.py
```diff
@@ -272,10 +279,7 @@ def impl(a, initial=None):
 @expose_and_overload(np.mean)
 @extending.overload_method(ArrayPointer, 'mean')
 def omnisci_array_mean(x):
-    if isinstance(x.eltype, types.Integer):
-        zero_value = 0
-    elif isinstance(x.eltype, types.Float):
-        zero_value = np.nan
+    zero_value = np.nan  # 0 if x.dtype is int
```
pearu, line 282: I think `mean` on integer input should return float value, that is, `mean` is a reduction op that promotes integers to float:
```suggestion
    zero_value = np.nan
```

rbc/tests/test_omnisci_array_methods.py
```diff
@@ -19,13 +22,27 @@ def omnisci():
...
+    ('max_empty_float32', (0, ), np.finfo(np.float32).min),
+    ('max_empty_float64', (0, ), np.finfo(np.float64).min),
     ('max_initial', (5, 4.0, 30.0), 30.0),
     ('mean', (5, 2.0), 2.0),
-    ('mean_empty_float', (0, ), np.nan),
-    ('mean_empty_int', (0, ), 0),
+    ('mean_empty_int8', (0, ), 0),
+    ('mean_empty_int16', (0, ), 0),
+    ('mean_empty_int32', (0, ), 0),
+    ('mean_empty_int64', (0, ), 0),
```
pearu, line 36: `mean` always returns `float`, so, mean of empty integer sequence ought to be `nan` as well. As in numpy:
```python
>>> import numpy
>>> numpy.mean(numpy.array([1, 2], dtype=numpy.int64))
1.5
>>> numpy.mean(numpy.array([], dtype=numpy.int64))
RuntimeWarning: Mean of empty slice.
nan
```

```suggestion
   ('mean', (5, 2), 2.0),
   ('mean_empty_int8', (0, ), np.nan),
   ('mean_empty_int16', (0, ), np.nan),
   ('mean_empty_int32', (0, ), np.nan),
   ('mean_empty_int64', (0, ), np.nan),
```

rbc/tests/test_omnisci_array_methods.py
```diff
@@ -19,13 +22,27 @@ def omnisci():
...
+    ('max_empty_int16', (0, ), np.iinfo(np.int16).min + 1),
+    ('max_empty_int32', (0, ), np.iinfo(np.int32).min + 1),
+    ('max_empty_int64', (0, ), np.iinfo(np.int64).min + 1),
+    ('max_empty_float32', (0, ), np.finfo(np.float32).min),
+    ('max_empty_float64', (0, ), np.finfo(np.float64).min),
     ('max_initial', (5, 4.0, 30.0), 30.0),
     ('mean', (5, 2.0), 2.0),
-    ('mean_empty_float', (0, ), np.nan),
-    ('mean_empty_int', (0, ), 0),
+    ('mean_empty_int8', (0, ), 0),
```
pearu, line 33: Hmm, according to `omnisci_array_mean`, the result ought to be `np.nan`. Is `mean_empty_int8` actually tested?

pearu (changes requested):
Let's follow numpy.mean conventions:
1. mean on integer input returns float value
2. mean on empty input returns always nan

rbc/tests/test_omnisci_array_methods.py
guilhermeleobas: Yes! Because the return type is an integer type, llvm cast np.nan to an integer value, which happens to be 0

pearu: ok, so when `mean` will return a float, we'll have `nan` here.

guilhermeleobas: exactly

rbc/tests/test_omnisci_array_methods.py
```diff
@@ -19,13 +22,27 @@ def omnisci():
...
+    ('max_empty_int16', (0, ), np.iinfo(np.int16).min + 1),
+    ('max_empty_int32', (0, ), np.iinfo(np.int32).min + 1),
+    ('max_empty_int64', (0, ), np.iinfo(np.int64).min + 1),
+    ('max_empty_float32', (0, ), np.finfo(np.float32).min),
+    ('max_empty_float64', (0, ), np.finfo(np.float64).min),
     ('max_initial', (5, 4.0, 30.0), 30.0),
-    ('mean', (5, 2.0), 2.0),
-    ('mean_empty_float', (0, ), np.nan),
-    ('mean_empty_int', (0, ), 0),
+    ('mean', (5, 2), 2.0),
```
pearu, line 32: Nit: I meant to keep both `2.0` and `2` variants of this `mean` test so that both float and int argument cases are covered:
```suggestion
    ('mean', (5, 2), 2.0),
    ('mean', (5, 2.0), 2.0),
```

pearu (approved):

Issue #377, opened 2021-11-10
Allocating arrays in the omnisci backend leaks memory


antocuni:
Consider the following code:
```python
from rbc.omnisci_backend import Array
@omnisci('float64(int32)')
def foo(size):
    a = Array(size, types.float64)
    return 1
```
And the corresponding IR:
```llvm
define double @foo_daiA(i32 %.1) local_unnamed_addr {
entry:
  %.5.i = zext i32 %.1 to i64
  %.6.i = tail call i8* @allocate_varlen_buffer(i64 %.5.i, i64 8), !noalias !1
  ret double 1.000000e+00
}
```

You can clearly see that the function leaks memory because the result of `allocate_varlen_buffer` is never freed.
In theory, we have code which is supposed to insert the appropriate calls to `free`, but for various reason it doesn't work.

The reason number 0 is that we don't have any test for this behavior: there are some tests which trigger the code in `omnisci_pipeline.py`, but we never check that the call to `free` is actually inserted in the produced IR. Probably this logic worked in the past and then broke at some point.

Apart from missing test, the more concrete reason for why this doesn't work is here:
https://github.com/xnd-project/rbc/blob/eb3bcafc3abd3a08635ed43d4908decd9d676538/rbc/omnisci_backend/omnisci_pipeline.py#L21-L35
The point of the logic above is to run the pass only if `Array` (or `Column`, `Buffer`, etc.) is ever referenced by the code. But it only checks for references through a `ir.FreeVar`: this happens only when we are compiling a nested function and the `Array` name is imported in the outer function. In the code above, `Array` is a `ir.Global`, because the `import` statement is done at global level.
This alone is a problem which should be solved, and it's likely that we never noticed because writing nested functions is exactly what you do in tests.

As a workaround, we can put `foo` inside a nested function:
```python
def make_foo():
    # by putting the import here, we make sure that Array is a FreeVar (look at the output of dis.dis)
    from rbc.omnisci_backend import Array
    @omnisci('float64(int32)')
    def foo(size):
        a = Array(size, types.float64)
        return 1
    return foo

foo = make_foo()
```
With this, the `FreeOmnisciBuffer` pass is actually executed (checked by putting `prints()`) but the end result is the same:
```llvm
define double @foo_daiA(i32 %.1) local_unnamed_addr {
entry:
  %.5.i = zext i32 %.1 to i64
  %.6.i = tail call i8* @allocate_varlen_buffer(i64 %.5.i, i64 8), !noalias !1
[...]

PR #392, opened 2021-12-22
Fix memory leaks when allocating buffers and arrays

antocuni:
I think this is ready for review, assuming all tests ill be green.
The core of this PR is to introduce the method `BufferPointer.free()`, which you can use to manually free the memory allocated by arrays and avoid the memory leak described in #377 .

Summary of changes:
- introduce `BufferPointer.free()` and finally remove the `@xfail` from `test_array_constructor_noreturn`.
- rename `free_omnisci_buffer` into `free_all_other_buffers` and add a docstring: the original name was very confusing IMHO, and moreover nowadays it's no longer omnisci-specific
- similarly, rename the llvm pass `FreeOmnisciBuffer` into `AutoFreeBuffers` and add a docstring to explain how it works (although it's probably broken)

Once this is merged, I plan to do a follow-up PR to finally fix `AutoFreeBuffers` and/or kill it if we decide that it is ultimately unfixable.

**EDIT**: I spoke too fast and the build is failing. Failures look unrelated but I'm going to investigate

**EDIT 2**: I re-ran the workflow and the build is all green now. It looks like we have flaky tests, which is a problem on its own but not for this PR :). @pearu @guilhermeleobas @brenocfg I think it's ready to review for real now

### DIFF
--- a/rbc/omnisci_backend/omnisci_buffer.py
+++ b/rbc/omnisci_backend/omnisci_buffer.py
@@ -258,9 +258,8 @@ def free_buffer(typingctx, buf):
         free_buffer_fn = irutils.get_or_insert_function(
             builder.module, free_buffer_fnty, free_buffer_fn_name)
 
-        # free all the buffers apart value_to_keep_alive
         [buf] = args
-        buf_ptr = builder.load(builder.gep(buf, [int32_t(0), int32_t(0)])) # buf.ptr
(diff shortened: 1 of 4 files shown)


pearu (approved):
LGTM, thanks @antocuni !

Looking forward to follow-up PRs. We'll need to test this feature on actual omnisci buffers as well.

guilhermeleobas (approved):
LGTM, thanks Antonio!

brenocfg (approved):
LGTM, nothing to add :). Thanks @antocuni!

PR #393, opened 2022-01-04
Maintainance of testing against omniscidb master

pearu:
As in the title.

Testing against CUDA enabled omniscidb requires https://github.com/omnisci/omniscidb-internal/pull/6187


guilhermeleobas (approved):
LGTM

PR #403, opened 2022-01-20
ENH: forbid direct calls to `RemoteOmnisci`

### LINKED ISSUE #400: Calling omnisci decorated function fails with cryptic error message
Consider:
```python
>>> from rbc.omniscidb import RemoteOmnisci
>>> omni = RemoteOmnisci()
>>> @omni('double(double)')
... def farenheight2celcius(f):
...     return (f - 32)* 5 / 9
... 
>>> _, r=omni.sql_execute('select farenheight2celcius(40)')
>>> print(list(r))
[(4.444444444444445,)]
>>> farenheight2celcius(40)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/pearu/git/xnd-project/rbc/rbc/remotejit.py", line 365, in __call__
    raise TypeError(
TypeError: specifying device is required when target has more than one device. Available devices: cpu, gpu
```

While calling a omnisci decorated function from Python is supposed to fail, users of rbc will likely still try to call the omnisci decorated functions from Python. The exception message should explain why it fails and what is the correct way to call the corresponding SQL operator in a query to omniscidb server.

tupui:
Closes #400

Add `__call__` to `RemoteOmnisci` which raises with clear documentation on how to properly use it.

### DIFF
--- a/rbc/omniscidb.py
+++ b/rbc/omniscidb.py
@@ -278,42 +278,11 @@ class RemoteOmnisci(RemoteJIT):
         self.has_cuda = None
         self._null_values = dict()
 
+        # Registered functions can only be called from a SQL query
+        self._supports_callable_caller = False
+
         # An user-defined device-LLVM IR mapping.
         self.user_defined_llvm_ir = {}
-
-    def __call__(self, *args, **kwargs):
-        msg = """Cannot call functions decorated by `RemoteOmnisci`.
-        These functions are registered on the database and can be called only
-        from a SQL query.
-
-        Following is a proper usage of `RemoteOmnisci`.
-
-        >>> from rbc.omniscidb import RemoteOmnisci
-        >>> omni = RemoteOmnisci()
-        >>> @omni('double(double)')
-        ... def farenheight2celcius(f):
-        ...     return (f - 32) * 5 / 9
-        ...
-        >>> _, r = omni.sql_execute('select farenheight2celcius(40)')
-        >>> print(list(r))
-        [(4.444444444444445,)]
-
-        Alternatively, you can define a Python function and decorate it
-        in a second step. This allows you to be able to call the Python
-        function. In any case, the decorated function cannot be called.
-
-        >>> omni = RemoteOmnisci()
-        >>> def farenheight2celcius(f):
-        ...     return (f - 32) * 5 / 9
-        ...
-        >>> decorated = omni('double(double)')(farenheight2celcius)
-        >>> _, r=omni.sql_execute('select farenheight2celcius(40)')
-        >>> print(list(r))
-        [(4.444444444444445,)]
-        >>> print(farenheight2celcius(40))
-        4.444444444444445
-        """
-        raise UnsupportedError(msg)
 
     def _init_thrift_typemap(self):
         """Initialize thrift type map using client thrift configuration.
--- a/rbc/omniscidb.py
+++ b/rbc/omniscidb.py
@@ -18,7 +18,7 @@ from .omnisci_backend import (
     BufferMeta, OmnisciColumnListType, OmnisciTableFunctionManagerType)
 from .targetinfo import TargetInfo
 from .irtools import compile_to_LLVM
-from .errors import ForbiddenNameError, OmnisciServerError
+from .errors import ForbiddenNameError, OmnisciServerError, UnsupportedError
 from .utils import parse_version
 from . import ctools
 from . import typesystem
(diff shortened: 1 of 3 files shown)


rbc/omniscidb.py
```diff
@@ -281,6 +281,40 @@ def __init__(self,
         # An user-defined device-LLVM IR mapping.
         self.user_defined_llvm_ir = {}
 
+    def __call__(self, *args, **kwargs):
+        msg = """Decorated functions called directly. Decorated function are
+        not meant to be called directly, but are registered and later
+        executed directly on the database.
+
```
antocuni, line 288: Just a suggestion, but what about the following?
```
Cannot call functions decorated by `RemoteOmnisci`.  These functions are registered on the database and can be called only from a SQL query.
```

rbc/omniscidb.py
```diff
@@ -281,6 +281,40 @@ def __init__(self,
...
+        ...
+        >>> decorated = omni('double(double)')(farenheight2celcius)
+        >>> _, r=omni.sql_execute('select farenheight2celcius(40)')
+        >>> print(list(r))
+        [(4.444444444444445,)]
+        >>> print(farenheight2celcius(40))
+        4.444444444444445
+        """
+        raise UnsupportedError(msg)
+
```
antocuni, line 317: wow, this is a a very good docstring :).
However, I'm not sure that it's good to have such a big and long error message. Maybe it's better to put this text in some online docs and put a link to it in the error message?

rbc/omniscidb.py
tupui: Sure, thanks for the suggestion.

rbc/omniscidb.py
tupui: Yes I can remove the good practice part. The initial thought was to ship this and then wait for the new doc to then link to it. Or can I already update some existing doc and link to it?

antocuni: ah ok, this approach makes sense, indeed

tupui:
Actually, shouldn't this be done on the `Caller` object instead? Because the function when decorated becomes a `Caller`. I think this is also why the notebook is failing.

pearu:
> Actually, shouldn't this be done on the `Caller` object instead? Because the function when decorated becomes a `Caller`. I think this is also why the notebook is failing.

Yes. We have:
```
omni = RemoteOmnisci()      # RemoteOmnisci instance
omni('double(double)')      # is Signature instance
omni('double(double)')(foo) # is Caller instance
```

tupui:
> Yes. We have:
> 
> ```
> omni = RemoteOmnisci()      # RemoteOmnisci instance
> omni('double(double)')      # is Signature instance
> omni('double(double)')(foo) # is Caller instance
> ```

Ok but then someone could do `Caller(x, device='cpu')` no?

pearu:
> Ok but then someone could do `Caller(x, device='cpu')` no? It should just do the computation on the remote instead of locally no?

Yes. `Caller` instance is callable and in the RemoteJIT case one can indeed execute functions remotely.

In the case of omnisci decorated functions, one would need to use `.local` Caller instance. For instance,
```
>>> farenheight2celcius.local(40)
```
should actually work but it does not atm. This is another issue that is closely related to https://github.com/xnd-project/rbc/issues/404

tupui:
> > Ok but then someone could do `Caller(x, device='cpu')` no? It should just do the computation on the remote instead of locally no?
> 
> Yes. `Caller` instance is callable and in the RemoteJIT case one can indeed execute functions remotely.
> 
> In the case of omnisci decorated functions, one would need to use `.local` Caller instance. For instance,
> 
> ```
> >>> farenheight2celcius.local(40)
> ```
> 
> should actually work but it does not atm. This is another issue that is closely related to #404

Oh right. But then it makes sense to use this `local` if the behaviour could be different for other cases.

I propose to complement the current error by direct the user to this feature (+ the other solution on how to not overwrite the Python function).

tupui:
I added an attribute `_locally_callable_caller` to `RemoteJIT`. This way the `Caller.__call__` can check this attribute and raise if it's called directly although it should not.

rbc/remotejit.py
```diff
@@ -358,6 +359,15 @@ def get_signatures(self):
     def __call__(self, *arguments, **options):
         """Return the result of a remote JIT compiled function call.
         """
+        if not self.remotejit._supports_callable_caller:
+            msg = (
+                "Cannot call functions decorated by "
+                f"{self.remotejit.__class__}. These functions are "
+                "registered on the database and can be called "
+                "only from a SQL query."
```
pearu, line 367: Nit: in general, there may exist other reasons for not supporting callable callers, and derivates of RemoteJIT may not be related to database applications. So, I suggest a more general exception message for the remotejit.py code:
```suggestion
                f"{type(self.remotejit)}."
```
and document this restriction in Omnisci related docs (to be created).

If we really want to add some explanation as above (which would be a nice feature for users not reading the docs) then the omniscidb.py could define, say, `OmnisciCaller` class derived from `Caller` with `__call__` method disabled and attach this to RemoteOmnisci class as `_caller_cls` attribute and let `RemoteJIT` use it whenever a `Caller` instance needs to be created. Let's come back to this when it is obvious that new users may hit this use case more often than expected.

pearu (changes requested):
LGTM! Thanks, @tupui!

I have a nit regarding the error message to detach omniscidb related matters from a more general remotejit.py code.

rbc/remotejit.py
tupui: I did the change.

For possible future improvement with `OmnisciCaller`, I think a factory pattern would work good here. We could have factories for all these different classes.

rbc/remotejit.py
```diff
@@ -316,6 +317,13 @@ def local(self):
         """Return Caller instance that executes function calls on the local
         host. Useful for debugging.
         """
+        if not self.remotejit._supports_callable_caller:
+            msg = (
+                "Cannot create a local `Caller` when using "
+                f"{type(self.remotejit)}."
```
tupui, line 323:
```suggestion
                f"{type(self.remotejit).__name__}."
```

rbc/remotejit.py
```diff
@@ -358,6 +366,13 @@ def get_signatures(self):
     def __call__(self, *arguments, **options):
         """Return the result of a remote JIT compiled function call.
         """
+        if not self.remotejit._supports_callable_caller:
+            msg = (
+                "Cannot call functions decorated by "
+                f"{type(self.remotejit)}."
```
tupui, line 372:
```suggestion
                f"{type(self.remotejit).__name__}."
```

pearu (approved):
LGTM! Thanks, @tupui!

PR #405, opened 2022-01-20
Implement dtype attribute access and comparison

guilhermeleobas:
As title. This PR implements access to `dtype` attribute for columns and arrays as well as an optimization which prunes dtype comparison branches at compile time.

### DIFF
--- a/rbc/tests/test_omnisci_column_basic.py
+++ b/rbc/tests/test_omnisci_column_basic.py
@@ -814,3 +814,36 @@ def test_column_different_sizes(omnisci):
 
     expected = [(10.0,), (60.0,), (180.0,), (240.0,), (300.0,)]
     assert list(result) == expected
+
+
+def test_column_dtype(omnisci):
+    from numba import types
+    table = omnisci.table_name
+
+    @omnisci('int32(Column<T>, RowMultiplier, OutputColumn<T>)',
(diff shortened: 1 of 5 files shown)


rbc/tests/test_omnisci_column_basic.py
```diff
@@ -550,6 +550,10 @@ def test_column_aggregate(omnisci, prop, oper):
     omnisci.reset()
     omnisci.register()
 
+
+    if prop == 'groupby' and oper == 'corr':
+        pytest.skip('omniscidb server crashes')
```
pearu, line 554: This is likely omniscidb version dependent. Could you create an issue about the crash and related details what are available?

pearu (changes requested):
Nice work @guilhermeleobas !

Could you also add a test where the `.dtype` is used as an input dtype argument. For instance:
```python
@omnisci('T[](T[])', T=['int32', 'int64'])
def zeros_like(x):
    z = zeros(x.size, x.dtype)
    return z
```

rbc/tests/test_omnisci_column_basic.py
guilhermeleobas: I've seen this failure in other builds. It seems to be a random thing.

pearu (approved):
LGTM when CI will be green. Thanks, @guilhermeleobas !

PR #424, opened 2022-01-31
Support remote calling of run-time UDF/UDTFs from Python

pearu:
As in the title.

Too many new RBC users (current count is 2) have fallen into the trap of calling omnisci decorated functions from Python.

So, this PR now implements basic support for calling UDFs and UDTFs from Python. Example:

```python
In [1]: from rbc.omniscidb import RemoteOmnisci, get_client_config

In [2]: from rbc.externals.omniscidb import set_output_row_size

In [3]: omnisci = RemoteOmnisci(**get_client_config())

In [4]: @omnisci('UDTF(int32, OutputColumn<int64> x)')
   ...: def arange(size, x):
   ...:     set_output_row_size(size)
   ...:     for i in range(size):
   ...:         x[i] = i
   ...:     return size
   ...: 

In [5]: arange
Out[5]: Caller(<function arange at 0x7f31aaa03b70>, Signature('UDTF(int32, OutputColumn<int64> x)'))

In [6]: print(arange)
arange['UDTF(int32, OutputColumn<int64> x)']

In [7]: arange(5)
Out[7]: OmnisciQueryCapsule('SELECT x FROM TABLE(arange(CAST(5 AS INT)))')

In [8]: arange(5).execute()
Out[8]: 
rec.array([(0,), (1,), (2,), (3,), (4,)],
          dtype=[('x', '<i8')])
In [15]: arange(5)['x']
Out[15]: array([0, 1, 2, 3, 4])

In [16]: arange(5.5)
<snip>
TypeError: found no matching function signature to given argument types:
    (float64) -> ...
  available function signatures:
    (int32) -> (Column<int64> x)
    - UDTF(int32, OutputColumn<int64> x)
```

Notice that `arange(5)` returns a `OmnisciQueryCapsule` instance that can be executed using `execute()` method. In the return, the output columns are collected into a numpy.recarray instance.

By default, the remote execution is postponed because UDTF calls can be composed. For example:

```python
In [17]: @omnisci('int32(Column<int64> x, OutputColumn<int64> y)')
    ...: def aincr(x, y):
    ...:     set_output_row_size(len(x))
    ...:     for i in range(len(x)):
    ...:         y[i] = x[i] + 1
    ...:     return len(y)
    ...: 

In [18]: print(aincr(arange(5)))
SELECT y FROM TABLE(aincr(CURSOR(SELECT x FROM TABLE(arange(CAST(5 AS INT))))))

In [19]: aincr(arange(5))['y']
Out[19]: array([1, 2, 3, 4, 5])

In [20]: aincr(aincr(arange(5)))['y']
Out[20]: array([2, 3, 4, 5, 6])
```

In the case of UDFs, the remote execution ~is immediate (unless one explicitly holds it back)~ is postponed as well:
```python
In [21]: @omnisci('int64(int64)')
    ...: def incr(x): return x + 1

In [22]: print(incr)
incr['int64(int64)']

In [23]: print(incr(5))
SELECT incr(CAST(5 AS BIGINT))

In [24]: incr(5).execute()
Out[24]: 6
```

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -21,7 +21,7 @@ from rbc import __version__
 # -- Project information -----------------------------------------------------
 
 project = 'RBC'
-copyright = '2022-%s, Xnd-Project Developers' % datetime.datetime.now().year
+copyright = '2018-%s, Xnd-Project Developers' % datetime.datetime.now().year
 author = 'Xnd-Project Developers'
 
 # The version info for the project you're documenting, acts as replacement for
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -21,7 +21,7 @@ from rbc import __version__
 # -- Project information -----------------------------------------------------
 
 project = 'RBC'
-copyright = '2018-%s, Xnd-Project Developers' % datetime.datetime.now().year
+copyright = '2022-%s, Xnd-Project Developers' % datetime.datetime.now().year
 author = 'Xnd-Project Developers'
 
 # The version info for the project you're documenting, acts as replacement for
--- a/rbc/omniscidb.py
+++ b/rbc/omniscidb.py
@@ -9,7 +9,7 @@ import warnings
 import configparser
 import numpy
 from collections import defaultdict, namedtuple
-from .remotejit import RemoteJIT
+from .remotejit import RemoteJIT, RemoteCallCapsule
 from .thrift.utils import resolve_includes
 from . import omnisci_backend
 from .omnisci_backend import (
@@ -19,7 +19,7 @@ from .omnisci_backend import (
 from .targetinfo import TargetInfo
 from .irtools import compile_to_LLVM
 from .errors import ForbiddenNameError, OmnisciServerError
-from .utils import parse_version
+from .utils import parse_version, DEFAULT, UNSPECIFIED
 from . import ctools
 from . import typesystem
 
@@ -212,6 +212,98 @@ def get_client_config(**config):
     return config
 
 
+def is_udtf(sig):
+    """Check if signature is a table function signature.
+    """
+    if sig[0].annotation().get('kind') == 'UDTF':
+        return True
+    for a in sig[1]:
+        if isinstance(a, (OmnisciOutputColumnType, OmnisciColumnType,
+                          OmnisciColumnListType, OmnisciTableFunctionManagerType)):
+            return True
+    return False
+
+
+output_buffer_sizer_map = dict(
+    ConstantParameter='kUserSpecifiedConstantParameter',
+    RowMultiplier='kUserSpecifiedRowMultiplier',
+    Constant='kConstant',
+    SpecifiedParameter='kTableFunctionSpecifiedParameter',
+    PreFlight='kPreFlightParameter')
+output_buffer_sizer_map[None] = output_buffer_sizer_map['RowMultiplier']
+
+user_specified_output_buffer_sizers = {
+    'kUserSpecifiedConstantParameter', 'kUserSpecifiedRowMultiplier',
+}
+
+
+def type_to_type_name(typ: typesystem.Type):
+    """Return typesystem.Type as DatumType name.
+    """
+    styp = typ.tostring(use_annotation=False)
+    type_name = dict(
+        int8='TINYINT',
+        int16='SMALLINT',
+        int32='INT',
+        int64='BIGINT',
+        float32='FLOAT',
+        float64='DOUBLE',
+    ).get(styp)
+    if type_name is not None:
+        return type_name
+    raise NotImplementedError(f'converting `{styp}` to DatumType not supported')
+
+
+def type_name_to_dtype(type_name):
+    """Return DatumType name as the corresponding numpy dtype.
+    """
+    dtype = dict(
+        SMALLINT=numpy.int16,
+        INT=numpy.int32,
+        BIGINT=numpy.int64,
+        FLOAT=numpy.float32,
+        # DECIMAL=,
+        DOUBLE=numpy.float32,
+        STR=numpy.str0,
+        # TIME=,
+        # TIMESTAMP=,
+        # DATE=,
+        BOOL=numpy.bool8,
+        # INTERVAL_DAY_TIME=,
+        # INTERVAL_YEAR_MONTH=,
+        # POINT=,
+        # LINESTRING=,
+        # POLYGON=,
+        # MULTIPOLYGON=,
+        TINYINT=numpy.int8,
+        # GEOMETRY=,
+        # GEOGRAPHY=
+    ).get(type_name)
+    if dtype is not None:
+        return dtype
+    raise NotImplementedError(
+        f'convert DatumType `{type_name}` to numpy dtype')
+
+
+class Query(RemoteCallCapsule):
+
+    def __call__(self, hold=DEFAULT, try_run=DEFAULT):
+        if is_udtf(self.ftype):
+            if hold is DEFAULT:
+                hold = True  # UDTFs executions are postponed
+            if try_run is DEFAULT:
+                try_run = False
+        else:
+            if hold is DEFAULT:
+                hold = False  # UDFs are evaluated immidiately
+            if try_run is DEFAULT:
+                try_run = False
+        return RemoteCallCapsule.__call__(self, hold=hold and not try_run, try_run=try_run)
+
+    def __getitem__(self, key):
+        return self.execute()[key]
+
+
 class RemoteOmnisci(RemoteJIT):
 
     """Usage:
@@ -245,7 +337,10 @@ class RemoteOmnisci(RemoteJIT):
         ColumnList='OmnisciColumnListType',
         TextEncodingDict='OmnisciTextEncodingDictType',
         TableFunctionManager='OmnisciTableFunctionManagerType<>',
+        UDTF='int32|kind=UDTF'
     )
+
+    remote_call_capsule_cls = Query
 
     def __init__(self,
                  user='admin',
@@ -279,7 +374,8 @@ class RemoteOmnisci(RemoteJIT):
         self._null_values = dict()
 
         # Registered functions can only be called from a SQL query
-        self._supports_callable_caller = False
+        self._supports_callable_caller = True
+        self._supports_local_caller = False
 
         # An user-defined device-LLVM IR mapping.
         self.user_defined_llvm_ir = {}
@@ -292,6 +388,7 @@ class RemoteOmnisci(RemoteJIT):
             if hasattr(typ, '_NAMES_TO_VALUES'):
                 for name, value in typ._NAMES_TO_VALUES.items():
                     typemap[typename][name] = value
+                    typemap[typename+'-inverse'][value] = name
 
     @property
     def version(self):
@@ -620,10 +717,13 @@ class RemoteOmnisci(RemoteJIT):
         result = self.thrift_call(
             'sql_execute', self.session_id, query, columnar, "", -1, -1)
 
-        Description = namedtuple("Description", ["name", "type_code", "null_ok"])
+        type_code_to_type_name = self.thrift_typemap['TDatumType-inverse']
+        Description = namedtuple("Description", ["name", "type_name", "null_ok"])
         descr = []
         for col in result.row_set.row_desc:
-            descr.append(Description(col.col_name, col.col_type.type, col.col_type.nullable))
+            descr.append(Description(col.col_name,
+                                     type_code_to_type_name[col.col_type.type],
+                                     col.col_type.nullable))
         return descr, self._make_row_results_set(result)
 
     _ext_arguments_map = None
@@ -874,11 +974,6 @@ class RemoteOmnisci(RemoteJIT):
                 'omniscidb 5.4 or newer, currently '
                 'connected to ', v)
         thrift = self.thrift_client.thrift
-        sizer_map = dict(
-            ConstantParameter='kUserSpecifiedConstantParameter',
-            RowMultiplier='kUserSpecifiedRowMultiplier',
-            Constant='kConstant',
-            SpecifiedParameter='kTableFunctionSpecifiedParameter')
 
         unspecified = object()
         inputArgTypes = []
@@ -900,7 +995,7 @@ class RemoteOmnisci(RemoteJIT):
                         'sizer argument must have type int32')
                 if _sizer is None:
                     _sizer = 'RowMultiplier'
-                _sizer = sizer_map[_sizer]
+                _sizer = output_buffer_sizer_map[_sizer]
                 # cannot have multiple sizer arguments
                 assert sizer_index == -1
                 sizer_index = consumed_index + 1
@@ -1012,12 +1107,6 @@ class RemoteOmnisci(RemoteJIT):
     def _register(self):
         if self.have_last_compile:
             return
-
-        def is_udtf(sig):
-            for a in sig[1]:
-                if isinstance(a, (OmnisciOutputColumnType, OmnisciColumnType)):
-                    return True
-            return False
 
         device_ir_map = {}
         llvm_function_names = []
@@ -1148,3 +1237,98 @@ class RemoteOmnisci(RemoteJIT):
                 print(f'compiler={compiler}')
             self._compiler = compiler
         return self._compiler
+
+    def caller_signature(self, signature: typesystem.Type):
+        """Return signature of a caller.
+
+        See RemoteJIT.caller_signature.__doc__.
+        """
+        if is_udtf(signature):
+            rtype = signature[0]
+            if not (rtype.is_int and rtype.bits == 32):
+                raise ValueError(
+                    f'UDTF implementation return type must be int32, got {rtype}')
+            rtypes = []
+            atypes = []
+            for atype in signature[1]:
+                _sizer = atype.annotation().get('sizer', UNSPECIFIED)
+                if _sizer is not UNSPECIFIED:
+                    if not (atype.is_int and atype.bits == 32):
+                        raise ValueError(
+                            f'sizer argument must have type int32, got {atype}')
+                    _sizer = output_buffer_sizer_map.get(_sizer, _sizer)
+                    if _sizer not in user_specified_output_buffer_sizers:
+                        continue
+                    atype.annotation(sizer=_sizer)
+                elif isinstance(atype, OmnisciTableFunctionManagerType):
+                    continue
+                elif isinstance(atype, OmnisciOutputColumnType):
+                    rtypes.append(atype.copy(OmnisciColumnType))
+                    continue
+                atypes.append(atype)
+            return typesystem.Type(*rtypes,
+                                   **dict(struct_is_tuple=True))(*atypes, **signature._params)
+        return signature
+
+    def get_types(self, *values):
+        """Convert values to the corresponding typesystem types.
+
+        See RemoteJIT.get_types.__doc__.
+        """
+        types = []
+        for value in values:
+            if isinstance(value, RemoteCallCapsule):
+                typ = value.__typesystem_type__
+                if typ.is_struct and typ._params.get('struct_is_tuple'):
+                    types.extend(typ)
+                else:
+                    types.append(typ)
+            else:
+                types.append(typesystem.Type.fromvalue(value))
+        return tuple(types)
+
+    # We define remote_compile and remote_call for Caller.__call__ method.
+    def remote_compile(self, func, ftype: typesystem.Type, target_info: TargetInfo):
+        """Remote compile function and signatures to machine code.
+
+        See RemoteJIT.remote_compile.__doc__.
+        """
+        if self.query_requires_register(func.__name__):
+            self.register()
+
+    def remote_call(self, func, ftype: typesystem.Type, arguments: tuple, try_run=False):
+        """
+        See RemoteJIT.remote_call.__doc__.
+        """
+        sig = self.caller_signature(ftype)
+        assert len(arguments) == len(sig[1]), (sig, arguments)
+        rtype = sig[0]
+        args = []
+        for a, atype in zip(arguments, sig[1]):
+            if isinstance(atype, (OmnisciColumnType, OmnisciColumnListType)):
+                if isinstance(a, RemoteCallCapsule):
+                    a = a.execute(try_run=True)
+                args.append(f'CURSOR({a})')
+            else:
+                args.append(f'CAST({a} AS {type_to_type_name(atype)})')
+        args = ', '.join(args)
+        is_udtf_call = is_udtf(ftype)
+        if is_udtf_call:
+            colnames = []
+            if rtype.is_struct and rtype._params.get('struct_is_tuple'):
+                for i, t in enumerate(rtype):
+                    n = t.annotation().get('name', f'out{i}')
+                    colnames.append(n)
+            else:
+                colnames.append(rtype.annotation().get('name', '*'))
+            q = f'SELECT {", ".join(colnames)} FROM TABLE({func.__name__}({args}))'
+        else:
+            q = f'SELECT {func.__name__}({args})'
+        if try_run:
+            return q
+        descrs, result = self.sql_execute(q + ';')
+        dtype = [(descr.name, type_name_to_dtype(descr.type_name)) for descr in descrs]
+        if is_udtf_call:
+            return numpy.array(list(result), dtype).view(numpy.recarray)
+        else:
+            return dtype[0][1](list(result)[0][0])
--- a/rbc/omniscidb.py
+++ b/rbc/omniscidb.py
@@ -13,13 +13,14 @@ from .remotejit import RemoteJIT, RemoteCallCapsule
 from .thrift.utils import resolve_includes
 from . import omnisci_backend
 from .omnisci_backend import (
+    OmnisciArrayType, OmnisciBytesType, OmnisciTextEncodingDictType,
     OmnisciOutputColumnType, OmnisciColumnType,
     OmnisciCompilerPipeline, OmnisciCursorType,
     BufferMeta, OmnisciColumnListType, OmnisciTableFunctionManagerType)
 from .targetinfo import TargetInfo
 from .irtools import compile_to_LLVM
 from .errors import ForbiddenNameError, OmnisciServerError
-from .utils import parse_version, UNSPECIFIED
+from .utils import parse_version
 from . import ctools
 from . import typesystem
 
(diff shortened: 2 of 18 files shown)


rbc/omniscidb.py
```diff
@@ -212,6 +212,98 @@ def get_client_config(**config):
...
+        if is_udtf(self.ftype):
+            if hold is DEFAULT:
+                hold = True  # UDTFs executions are postponed
+            if try_run is DEFAULT:
+                try_run = False
+        else:
+            if hold is DEFAULT:
+                hold = False  # UDFs are evaluated immidiately
+            if try_run is DEFAULT:
+                try_run = False
```
tupui, line 300: We can simplify this
```suggestion
        if try_run is DEFAULT:
            try_run = False
        if is_udtf(self.ftype) and hold is DEFAULT:
            hold = True  # UDTFs executions are postponed
        else:
            hold = False  # UDFs are evaluated immidiately
```

rbc/tests/test_omnisci.py
```diff
@@ -72,9 +73,7 @@ def test_direct_call(omnisci):
     def farenheight2celcius(f):
```
tupui, line 73: It's a general typo we should fix at some point: _fahrenheit_ is the correct orthography.

rbc/tests/test_omnisci_caller.py
```diff
@@ -0,0 +1,89 @@
...
+    def aincr(x, dx, y):
+        size = len(x)
+        set_output_row_size(size)
+        for i in range(size):
+            y[i] = x[i] + dx
+        return size
+
+
+def test_remote_udf_evaluation(omnisci):
+    myincr = omnisci.get_caller('myincr')
```
tupui, line 41: This mechanism is equivalent to an import in a Python world right?

tupui (commented):
Thanks @pearu. I have a few suggestions for now.

tupui:
I think the holding mechanism on UTF, UDTF behaviour might be confusing for a user. I think that printing should always hold and calling should call. Some user will not think about the distinction between UDF and UDTF and just think about it in terms of function. Which I think is the natural way of thinking about this when you first see this kind of construction.

pearu:
> I think the holding mechanism on UTF, UDTF behaviour might be confusing for a user. I think that printing should always hold and calling should call.

This applies to UDFs.

For UDTFs, however, calling should hold to enable the composition of UDTFs. For instance, in the
`aincr(arange(5))` case, the subexpression `arange(5)` should not be executed, otherwise, we would transfer the tables from server-to-client-to-server-to-client as opposite to server-to-client.

Btw, the feature implemented in this PR would not be the recommended way of calling UDF/UDTFs from Python as there exist more powerful tools for that. For instance, here we support only direct calls to UDF/UDTFs but SQL features like filtering, grouping, etc are not supported. I foresee the feature to be useful for quick experimentation and testing only.

tupui:
> > I think the holding mechanism on UTF, UDTF behaviour might be confusing for a user. I think that printing should always hold and calling should call.
> 
> This applies to UDFs.
> 
> For UDTFs, however, calling should hold to enable the composition of UDTFs. For instance, in the `aincr(arange(5))` case, the subexpression `arange(5)` should not be executed, otherwise, we would transfer the tables from server-to-client-to-server-to-client as opposite to server-to-client.

What I am proposing is to have by default `hold=True` for both UDF and UDTFs. This would be to try preventing the user to do this except for inspection purposes. So if they want a local execution, they would need to explicitly call `.execute()` otherwise it would print the SQL. Cf your other comment.

> Btw, the feature implemented in this PR would not be the recommended way of calling UDF/UDTFs from Python as there exist more powerful tools for that. For instance, here we support only direct calls to UDF/UDTFs but SQL features like filtering, grouping, etc are not supported. I foresee the feature to be useful for quick experimentation and testing only.

I will describe it that way in the documentation.

doc/conf.py
```diff
@@ -21,7 +21,7 @@
 # -- Project information -----------------------------------------------------
 
 project = 'RBC'
-copyright = '2018-%s, Xnd-Project Developers' % datetime.datetime.now().year
+copyright = '2022-%s, Xnd-Project Developers' % datetime.datetime.now().year
```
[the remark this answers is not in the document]
pearu: Fixed.

rbc/omniscidb.py
pearu: Fixed. Now also execution of UDFs is postponed by default.

rbc/tests/test_omnisci.py
pearu: Fixed.

rbc/tests/test_omnisci_caller.py
pearu: Not quite. The `get_caller` returns `CommonNameCallers` instance that may contain more than one Caller instance.

A user typically would not use `get_caller` method. In tests, we define UDF/UDTFs in a fixture and `get_caller` provides a way to retrieve the corresponding definitions from a test function.

pearu:
> What I am proposing is to have by default `hold=True` for both UDF and UDTFs.

This makes sense. Done. Thanks, @tupui !

rbc/tests/test_omnisci_caller.py
tupui: Oh I see, it gives back the undecorated function.

tupui (approved):
Thanks for the updates!

guilhermeleobas (commented):
Great work, Pearu!

I've done an initial review and have a few comments. Most of them are nits.

rbc/tests/test_omnisci_caller.py
```diff
@@ -0,0 +1,116 @@
...
+
+    assert_equal(arange(3, np.float32(1))['x'], np.arange(3, dtype=np.float32) + 1)
+    # The following test fails because SELECT seems to downcast
+    # float64 to float32. TODO: investigate.
+    # assert_equal(arange(3, np.float64(1))['x'], np.arange(3, dtype=np.float64) + 1)
+
+    assert_equal(arange(3, np.int64(1))['x'], np.arange(3, dtype=np.int64) + 1)
+    # The following test fails because SELECT seems to upcast int32 to
+    # int64. TODO: investigate.
+    # assert_equal(arange(3, np.int32(1))['x'], np.arange(3, dtype=np.int32) + 1)
```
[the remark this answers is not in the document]
pearu: The thing is that I am not sure if this is an omniscidb bug or a feature.

rbc/omniscidb.py
```diff
@@ -212,6 +212,95 @@ def get_client_config(**config):
...
+
+    def __call__(self, hold=DEFAULT, try_run=DEFAULT):
+        if try_run is DEFAULT:
+            try_run = False
+        if hold is DEFAULT:
+            # UDF/UDTFs executions are hold by default to support
+            # extension functions compositions in a data-transfer
+            # efficient way.
+            hold = True
+        return RemoteCallCapsule.__call__(self, hold=hold and not try_run, try_run=try_run)
```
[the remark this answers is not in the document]
pearu: Indeed, not anymore. Before, the default to `hold` was dependent on whether `self` holds a UDF or UDTF. Will fix it..

rbc/omniscidb.py
```diff
@@ -279,7 +371,8 @@ def __init__(self,
         self._null_values = dict()
 
         # Registered functions can only be called from a SQL query
-        self._supports_callable_caller = False
+        self._supports_callable_caller = True
```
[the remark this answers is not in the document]
pearu: yes. Are you suggesting removing `_supports_callable_caller`?

rbc/omniscidb.py
```diff
@@ -212,6 +212,95 @@ def get_client_config(**config):
...
+        # GEOMETRY=,
+        # GEOGRAPHY=
+    ).get(type_name)
+    if dtype is not None:
+        return dtype
+    raise NotImplementedError(
+        f'convert DatumType `{type_name}` to numpy dtype')
+
+
+class Query(RemoteCallCapsule):
```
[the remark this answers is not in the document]
pearu: It is a good suggestion. Since the object is exposed to end-users, it's str/repr representations ought to be informative:
Suggestion:
```python
>>> str(incr(5)))
'SELECT incr(CAST(5 AS BIGINT))'
>>> repr(incr(5))
OmniciQueryCapsule('SELECT incr(CAST(5 AS BIGINT))')
```
@guilhermeleobas @tupui what do you think?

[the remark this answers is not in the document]
tupui: I like the proposal.

Another question I am having is: do we want to cast the input? This might confuse some user no? If we have int to float it's fine but in some case we will have float to int and without any warning the result can introduce bug in users' code.

rbc/remotejit.py
```diff
@@ -371,24 +388,127 @@ def __call__(self, *arguments, **options):
...
+
+    def __repr__(self):
+        return f"{type(self).__name__}({self.name}, {self.callers})"
+
+    def __str__(self):
+        lst = []
+        for caller in self.callers:
+            lst.append(str(caller.signature))
+        sigs = '\n  '.join(lst)
+        return f'{self.name}:\n  {sigs}'
```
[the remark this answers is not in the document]
pearu: I think str result should represent the actual content of the `Signature`:
```
str(myincr) -> Signature(['T(T)', T=[...], ...])
str(myincr.normalized()) -> Signature(['int32(int32)', ...])
```
or similar.

The above is not ideal and I think a better approach is needed here because `myincr` is exposed to end-user.

rbc/remotejit.py
```diff
@@ -678,30 +817,38 @@ def remote_compile(self, func, ftype: Type, target_info: TargetInfo):
...
         """Call function remotely on given arguments.
 
         The input function `func` is called remotely by sending the
         arguments data to remote host where the previously compiled
         function (see `remote_compile` method) is applied to the
         arguments, and the result is returned to local process.
+
+        If `try_run` is True then return an object that specifies
+        remote call but does not execute it. The type of return object
+        is custom to particular RemoteJIT specialization.
```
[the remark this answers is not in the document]
pearu: `try_run` is partially supported in RemoteJIT.

The `try_run` feature is internally used in composition of remote functions in order to avoid transferring intermediate results between the client and the server. In this sense, both `try_run` and `delay_call` are a little misleading. Perhaps `postpone_execution` or similar?

The feature requires more work for RemoteJIT as currently the intermediate results are transferred..

rbc/remotejit.py
```diff
@@ -211,19 +211,36 @@ def best_match(self, func, atypes: tuple) -> Type:
...
+    def _format_available_function_types(self, func):
+        available_types = self.normalized(func).signatures
+        lines = []
+        for typ in available_types:
+            sig = self.remotejit.caller_signature(typ)
+            if sig == typ:
+                lines.append(f'{sig}')
+            else:
+                lines.append(f'{sig}\n  - {typ}')
+        return '    ' + '\n    '.join(lines)
```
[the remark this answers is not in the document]
pearu: I agree. The output here can be certainly improved.

rbc/remotejit.py
```diff
@@ -371,24 +388,127 @@ def __call__(self, *arguments, **options):
...
+        call.
+        """
+        return self.caller.remotejit.caller_signature(self.ftype)[0]
+
+    def __repr__(self):
+        return (f'{type(self).__name__}({self.caller!r}, {self.target_info},'
+                f' {self.ftype}, {self.arguments})')
+
+    def __str__(self):
+        return f'{type(self).__name__}({self(try_run=True)})'
```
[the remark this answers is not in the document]
pearu: Applied.

rbc/omniscidb.py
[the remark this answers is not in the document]
pearu: Applied.

rbc/omniscidb.py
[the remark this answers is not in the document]
pearu: Eliminated `_supports_callable_caller`.

rbc/omniscidb.py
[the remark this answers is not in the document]
pearu: I agree that some casts can be eliminated, but not all. For instance, `incr(np.int64(5))` should result in calling int64 version of `incr`, that is, `str(np.int64(5))` should lead to `'SELECT incr(CAST(5 AS BIGINT))'`.
On the other hand, `str(np.int32(5))` could be `SELECT incr(5)`.

Using CAST depends on how omniscidb interprets literal values and to drop some casts, RBC ought to record the casting rules of omniscidb (which are still in flux). Since we don't have this recorded, explicit CAST is better than no CAST.

rbc/remotejit.py
```diff
@@ -460,8 +582,10 @@ def __init__(self, host='localhost', port=11532,
         self._callers = []
         self._last_compile = None
         self._targets = None
-        # Some callers cannot be locally called
+        # Some callers cannot be called
         self._supports_callable_caller = True
```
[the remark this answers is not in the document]
pearu: `_supports_callable_caller` is eliminated.

rbc/remotejit.py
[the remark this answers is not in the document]
pearu: Fixed. See `test_udf/udtf_string_repr` in `tests/test_omnisci_caller.py` for updated behavior.

rbc/omniscidb.py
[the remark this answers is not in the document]
pearu: Query is now OmnisciQueryCapsule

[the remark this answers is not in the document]
pearu: > in some case we will have float to int and without any warning the result can introduce bug in users code ..

Not possible: the casts are generated based on user inputs and on available implementations. If there is a mismatch in input types and expected argument types, a TypeError will be raised. Try, for instance, `arange(5.5)` that will fail because `arange` expects integer input.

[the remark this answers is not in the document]
tupui: Oh great 👍

rbc/omniscidb.py
```diff
@@ -212,6 +212,95 @@ def get_client_config(**config):
...
+    return False
+
+
+output_buffer_sizer_map = dict(
+    ConstantParameter='kUserSpecifiedConstantParameter',
+    RowMultiplier='kUserSpecifiedRowMultiplier',
+    Constant='kConstant',
+    SpecifiedParameter='kTableFunctionSpecifiedParameter',
+    PreFlight='kPreFlightParameter')
+output_buffer_sizer_map[None] = output_buffer_sizer_map['RowMultiplier']
```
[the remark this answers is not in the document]
pearu: It is correct. The code is relevant when one specifies `int32 | sizer` which is equivalent to `int32 | sizer=None`. The default sizer is `RowMultiplier`.

`output_buffer_sizer_map.get(key, default_value)` gives default value when `key` is incorrectly specified (a typo, for instance) but it should fail.

rbc/tests/test_omnisci_caller.py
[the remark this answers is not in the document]
pearu: Done.

rbc/remotejit.py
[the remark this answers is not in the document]
pearu: Done.

rbc/remotejit.py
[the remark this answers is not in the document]
pearu: Using postpone_execution.

rbc/omniscidb.py
```diff
@@ -1148,3 +1243,157 @@ def compiler(self):
...
+            if is_udtf(typ):
+                return f'UDTF({", ".join(args)})'
+            else:
+                return f'({", ".join(args)}) -> {self.format_type(typ[0])}'
+        use_typename = False
+        if typ.is_struct and typ._params.get('struct_is_tuple'):
+            return f'({", ".join(map(self.format_type, typ))})'
+        if isinstance(typ, OmnisciOutputColumnType):
+            p = tuple(map(self.format_type, typ[0]))
+            typ = typesystem.Type(('OutputColumn',) + p, **typ._params)
```
tupui, line 1321: Shouldn't `typ` already be in the special type as an attribute?

rbc/remotejit.py
```diff
@@ -710,6 +849,22 @@ def python(self, statement):
...
+        ----------
+        ftype: Type
+          typesystem type of a function
+
+        Returns
+        -------
+        ftype: Type
+          typesystem type of a function with normalization hools applied
+        """
+        assert ftype.is_function, ftype
```
tupui, line 865: Not a fan of using `assert` for control flow but I guess none of our users would call Python with `-O`.

rbc/tests/test_omnisci_caller.py
```diff
@@ -0,0 +1,192 @@
...
+
+def test_remote_udf_typeerror(omnisci):
+    myincr = omnisci.get_caller('myincr')
+    try:
+        myincr("abc")
+    except TypeError as msg:
+        assert_equal(str(msg), '''\
+found no matching function type to given argument types:
+    string
+  available function types:
```
tupui, line 167: Noting this here. Not sure about `function types`, maybe we should talk about signature instead. Also it should say that it's the input type which is causing a problem. Because right now some users can be confused and not get why there is just one type and below there is a list of signatures

tupui (commented):
Just some thoughts for me.

rbc/omniscidb.py
pearu: The aim of this remapping is to apply `format_type` to the parameters of OmnisciOutputColumnType (as T is a parameter in `OutputColumn<T>`). Since `OmnisciOutputColumnType` is a custom `typesystem.Type` then `typesystem` has no information about its internal structure (for instance, parameters of a custom type can be arbitrary Python objects). So, here we explicitly apply `format_type` to the parameters (line 1320) and construct a temporary custom type (line 1321) that does the right thing when calling its `tostring`.

rbc/remotejit.py
pearu: It is not a control flow. I am using `assert` only for quick internal sanity checks. If somebody hits it, we have two options: either there is an internal bug that we'll need to fix, or the `assert` needs to be replaced with a proper exception raise as the user input can cause the hit.

rbc/tests/test_omnisci_caller.py
pearu: @tupui Are you suggesting the replacement: `function type` -> `signature type`? Or changing the whole exception message? Could you provide the desired message sample?

rbc/remotejit.py
tupui: ok I thought it was used for that because the input parameter is then just going through.

rbc/tests/test_omnisci_caller.py
tupui: Personally, I read the following better
```
found no matching function signature for given argument types:
     (string) -> ??
   available function signatures:
     (int32) -> int32
     (int64) -> int64
     (float32) -> float32
     (float64) -> float64
```

rbc/omniscidb.py
tupui: ok I am missing some pieces to fully get the whole logic 😅

rbc/remotejit.py
pearu: RemoteOmnisci redefines the method to handle UDTFs that have different signatures for the implementation functions and the corresponding SQL operators.

rbc/tests/test_omnisci_caller.py
pearu: Fixed. Using `(string) -> ...` though as question marks have negative energy ;)

tupui: haha, perfect thanks 👍

rbc/remotejit.py
tupui: Ah yes I see.

tupui:
I found a bug when you print and you have overloading:

```python
from rbc.omniscidb import RemoteOmnisci
omni = RemoteOmnisci(user='admin',
                     password='HyperInteractive',
                     host='localhost',
                     port=6274,
                     dbname='omnisci',)


@omni('double(double, double)')
def fahrenheit2celsius(f, offset):
    return offset + (f - 32) * 5 / 9


@omni('double(double)')
def fahrenheit2celsius(f):
    return (f - 32) * 5 / 9


# This is working fine
query = f'select fahrenheit2celsius(40)'
_, result = omni.sql_execute(query)
query = f'select fahrenheit2celsius(40, -1)'
_, result = omni.sql_execute(query)

# Printing not
str(fahrenheit2celsius(40, -1))

# Execute not as well
fahrenheit2celsius(40, -1).execute()
```

```python-traceback
E           TypeError: found no matching function signature to given argument types:
E               (float64, int64) -> ...
E             available function signatures:
E               (float64) -> float64
```

It only considers the last definition.

pearu:
> I found a bug when you print or execute and you have overloading:

Thanks for catching this! A fix is coming up but until then use:
```python
fahrenheit2celsius = omni.get_caller(''fahrenheit2celsius)
```

tupui:
> > I found a bug when you print or execute and you have overloading:
> 
> Thanks for catching this! ~A fix is coming up but until then use~:
> 
> ```python
> fahrenheit2celsius = omni.get_caller(''fahrenheit2celsius)
> ```
> 
> UPDATE: fixed.

Thanks for the quick fix, I confirm it works for me too.

tupui:
This one is really a nitpick and not sure we care. 

```python
query = str(fahrenheit2celsius(40))  # 'fahrenheit2celsius(CAST(40 AS DOUBLE))'
omni.sql_execute(query)
```

```python-traceback
rbc.errors.OmnisciServerError: Non-query expression encountered in illegal context
```

pearu:
> This one is really a nitpick and not sure we care. (Note that this crashes the server).

It makes sense to have the output of `str(fahrenheit2celsius(40))` be a valid SQL statement. Fixed.

> crashes the server

What does it mean here? The server raises an error message or the server dies?

tupui:
> > This one is really a nitpick and not sure we care. (Note that this crashes the server).
> 
> It makes sense to have the output of `str(fahrenheit2celsius(40))` be a valid SQL statement. Fixed.
> 
> > crashes the server
> 
> What does it mean here? The server raises an error message or the server dies?

It was dying after this statement
```
2022-02-03T13:26:20.243683 E 34 0 508 DBHandler.cpp:1328 SQL Error: Non-query expression encountered in illegal context
```

pearu:
> It was dying after this statement

I can reproduce the SQL error but I would not call it dying because the server stays alive and is functional.

guilhermeleobas:
Hi Pearu, about the casting. Shouldn't the last query cast the input to `TINY INT`?

```python
(Pdb++) myincr(3)
OmnisciQueryCapsule('SELECT myincr(CAST(3 AS BIGINT))')
(Pdb++) myincr(np.int32(3))
OmnisciQueryCapsule('SELECT myincr(CAST(3 AS INT))')
(Pdb++) myincr(np.int8(3))
OmnisciQueryCapsule('SELECT myincr(CAST(3 AS INT))')
```

tupui:
> > It was dying after this statement
> 
> 
> 
> I can reproduce the SQL error but I would not call it dying because the server stays alive and is functional.

Ah on my side it exited after the statement.

pearu:
> Shouldn't the last query cast the input to `TINY INT`?

No (unless also `int8 myincr(int8)` is defined). The SQL statement in `str` output is generated using the existing set of registered operators, that is, when calling `myincr(np.int8(3))`, a client-side argument matching is performed.

rbc/omniscidb.py
```diff
@@ -212,6 +213,111 @@ def get_client_config(**config):
...
+    """Return typesystem.Type as DatumType name.
+    """
+    styp = typ.tostring(use_annotation=False, use_name=False)
+    type_name = dict(
+        int8='TINYINT',
+        int16='SMALLINT',
+        int32='INT',
+        int64='BIGINT',
+        float32='FLOAT',
+        float64='DOUBLE',
```
guilhermeleobas, line 272: Add bytes (TextEncodingNone) here?

rbc/thrift/server.py
```diff
@@ -141,8 +142,11 @@ def run_bg(dispatcher, thrift_file, options, startup_time=5):
                 p.terminate()
             raise RuntimeError(
                 'failed to start up rpc_thrift server'
-                ' (was alive={}, startup_time={}s)'
-                .format(is_alive, startup_time))
+                f' (was alive={is_alive}, startup time={startup_time}s,'
+                f' elapsed time={time.time() - start})')
+        if number_of_tries > 1:
+            print(f'More than one try ({number_of_tries}) in starting up rpc_thrift_server.'
+                  f' Total elapsed time is {time.time() - start} seconds.')
```
guilhermeleobas, line 149:
```suggestion
            import warnings
            warnings.info(f'More than one try ({number_of_tries}) in starting up rpc_thrift_server.'
                         f' Total elapsed time is {time.time() - start} seconds.')
```

rbc/typesystem.py
```diff
@@ -1470,6 +1493,18 @@ def match(self, other):
...
+                return
+            elif self.is_custom:
+                if type(self) is not type(other):
+                    return
+                if self[0] == other[0]:
+                    return 0
+                # WARNING: if the state of custom type depends on
+                # params or annotations then it must re-define match
+                # method accordingly.
+                return
```
guilhermeleobas, line 1507: Emit a warning here?

guilhermeleobas (approved):
@pearu, I have a few more suggestions, but the PR looks good to me.

rbc/omniscidb.py
```diff
@@ -212,6 +213,111 @@ def get_client_config(**config):
...
+
+class OmnisciQueryCapsule(RemoteCallCapsule):
+
+    use_execute_cache = True
+
+    def __repr__(self):
+        return f'{type(self).__name__}({str(self)!r})'
+
+    def __getitem__(self, key):
+        return self.execute()[key]
```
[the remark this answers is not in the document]
pearu: Gone.

rbc/omniscidb.py
```diff
@@ -241,12 +347,19 @@ def add(a, b):
...
+        PreFlight='int32|sizer=PreFlight',
         ColumnList='OmnisciColumnListType',
         TextEncodingDict='OmnisciTextEncodingDictType',
         TableFunctionManager='OmnisciTableFunctionManagerType<>',
+        UDTF='int32|kind=UDTF'
     )
 
+    remote_call_capsule_cls = OmnisciQueryCapsule
+    default_remote_call_hold = True
+    supports_local_caller = False
```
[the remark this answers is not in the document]
pearu: yes, perhaps, but not in this PR if this is ok as it requires some analysis to see if mixins simplify the current code or not. Let's keep mixins in mind though...

rbc/omniscidb.py
```diff
@@ -212,6 +213,111 @@ def get_client_config(**config):
...
+
+def get_sizer_enum(t):
+    """Return sizer enum value as defined by the omniscidb server.
+    """
+    sizer = t.annotation()['sizer']
+    sizer = output_buffer_sizer_map.get(sizer or None, sizer)
+    for shortname, fullname in output_buffer_sizer_map.items():
+        if sizer == fullname:
+            return sizer
+    print(f'Warning: unknown sizer {sizer}')
```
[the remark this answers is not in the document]
pearu: Fixed, raising an exception now.

rbc/thrift/server.py
pearu: Fixed.

rbc/omniscidb.py
pearu: It turns out that TextEncodingNone is not exactly a TDatumType.
This code is relevant only in the UDTF context but we currently do not support `Column<Bytes>`.
However, I was able to add bytes support to UDF cases.

rbc/typesystem.py
pearu: Yes, makes sense.

pearu: Done.

PR #426, opened 2022-02-03
ENH: add unregister of callers locally and on the server

tupui:
Allow to clean the server from registered functions.

Also noted that when the function does not exist on the server, the message is not very clear to me:

```
rbc.errors.OmnisciServerError: From line 1, column 8 to line 1, column 32: No match found for function signature farenheight2celcius(<NUMERIC>)
```

### DIFF
--- a/rbc/omniscidb.py
+++ b/rbc/omniscidb.py
@@ -364,6 +364,14 @@ class RemoteOmnisci(RemoteJIT):
             m = re.match(r'.*Exception: (.*)', msg.error_msg)
             if m:
                 raise OmnisciServerError(f'{m.group(1)}')
+            m = re.match(
+                    r'(.*)\: No match found for function signature (.*)\(.*\)',
+                         msg.error_msg
+            )
+            if m:
+                msg = (f"Undefined function call {m.group(2)!r} in"
+                       f" SQL statement: {m.group(1)}")
+                raise OmnisciServerError(msg)
             m = re.match(r'.*SQL Error: (.*)', msg.error_msg)
             if m:
                 raise OmnisciServerError(f'{m.group(1)}')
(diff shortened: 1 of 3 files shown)


pearu:
> when the function does not exist on the server, the message is not very clear

I agree. We could rewrite the message by removing `From line 1, column 8 to line 1, column 32` part from the line as the line/column references correspond to SQL query line that is generated by rbc. On the other hand, this part should be kept if the SQL query line is defined by an user.

tupui:
> > when the function does not exist on the server, the message is not very clear
> 
> I agree. We could rewrite the message by removing `From line 1, column 8 to line 1, column 32` part from the line as the line/column references correspond to SQL query line that is generated by rbc. On the other hand, this part should be kept if the SQL query line is defined by an user.

Do you mean something like that then?

```
rbc.errors.OmnisciServerError: No match found for function signature farenheight2celcius(<NUMERIC>). Error in SQL statement: From line 1, column 8 to line 1, column 32
```

Or should I re-write it more like

```
rbc.errors.OmnisciServerError: Undefined function call farenheight2celcius in SQL statement: From line 1, column 8 to line 1, column 32
```

pearu:
The second option reads better to me.

rbc/omniscidb.py
```diff
@@ -364,6 +364,14 @@ def thrift_call(self, name, *args, **kwargs):
             m = re.match(r'.*Exception: (.*)', msg.error_msg)
             if m:
                 raise OmnisciServerError(f'{m.group(1)}')
+            m = re.match(
+                    r'(.*)\: No match found for function signature (.*)\(.*\)',
+                         msg.error_msg
+            )
```
tupui, line 477:
```suggestion
            m = re.match(
                r'(.*)\: No match found for function signature (.*)\(.*\)',
                msg.error_msg
            )
```

guilhermeleobas (approved):

PR #431, opened 2022-02-09
MAINT: define public api of omnisci_backend functions

tupui:
Define the public API of `rbc.omnisci_backend`. This is handy for exploration from an IDE.

The only issue is that there is no inline information about the doc due to the way overloading is done. It's also why I added a skip of the rule F822 (undefined name in `__all__`)


rbc/omnisci_backend/__init__.py
```diff
@@ -1,13 +1,16 @@
...
 from .omnisci_bytes import *  # noqa: F401, F403
 from .omnisci_metatype import *  # noqa: F401, F403
 from .omnisci_text_encoding import *  # noqa: F401, F403
-from .numpy_ufuncs import *  # noqa: F401, F403
 from .omnisci_pipeline import *  # noqa: F401, F403
-from .python_operators import *  # noqa: F401, F403
 from .omnisci_column_list import *  # noqa: F401, F403
 from .omnisci_table_function_manager import *  # noqa: F401, F403
+
+import rbc.omnisci_backend.mathimpl as math  # noqa: F401
```
guilhermeleobas, line 12: I think you can revert this one. Users don't use math from `rbc.omnisci_backend.mathimpl` but rather importing math directly:

```python
import math

@omnisci('float64(float64)')
def log2(x):
    return math.log2(x)
```

The same for python_operators module.

tupui: It would still work but I did this so that we don't pollute the env with what's inside these when we do `import rbc.omnisci_backend`.

guilhermeleobas:
@tupui, I've rebased your PR with the main branch to pull https://github.com/xnd-project/rbc/pull/433 code. Hopefully this should fix all the CI failures we're seeing.

tupui:
> @tupui, I've rebased your PR with the main branch to pull #433 code. Hopefully this should fix all the CI failures we're seeing.

Thanks @guilhermeleobas!

pearu (approved):
LGTM! Thanks, @tupui!

With @guilhermeleobas approval, the PR can be landed.

PR #435, opened 2022-02-14
add support for enumerate(Array, Column)

### LINKED ISSUE #420: Support `enumerate` for UD[T]F
Classic Python grammar should be supported. `enumerate` is not supported in UD[T]F.

guilhermeleobas:
Fix #420 

Heavily inspired in the `iternext` code for tuples:
https://github.com/numba/numba/blob/61ec1fd0f69aeadece218dccf4c39ebc5c7dfbc4/numba/cpython/tupleobj.py#L159-L187

### DIFF
--- a/.github/workflows/rbc_test.yml
+++ b/.github/workflows/rbc_test.yml
@@ -105,7 +105,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     timeout-minutes: 20
     strategy:
-      fail-fast: true
+      fail-fast: false
       matrix:
         os: [ubuntu-latest]
         python-version: ['3.9', '3.8', '3.7']
--- a/.github/workflows/rbc_test.yml
+++ b/.github/workflows/rbc_test.yml
@@ -35,7 +35,8 @@ jobs:
     name: ${{ matrix.os }} - Python v${{ matrix.python-version }} - Numba v${{ matrix.numba-version }}
     runs-on: ${{ matrix.os }}
     strategy:
-      fail-fast: true
+      # setting fail-fast=true seems to cause connection issues in remotejit tests
+      fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
         python-version: ['3.9', '3.8', '3.7']
--- a/rbc/tests/test_omnisci_column_basic.py
+++ b/rbc/tests/test_omnisci_column_basic.py
@@ -869,5 +869,6 @@ def test_column_enumerate(omnisci):
     _, result = omnisci.sql_execute(
         f'select * from table(col_enumerate(cursor(select i4 from {omnisci.table_name})))')
     _, expected_result = omnisci.sql_execute(
-        f'select i4 from {omnisci.table_name}')
-    assert list(result) == list(expected_result)
+        f'select rowid, i4 from {omnisci.table_name} order by rowid;')
+    for (r,), (_, e) in zip(list(result), list(expected_result)):
+        assert r == e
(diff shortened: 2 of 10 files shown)


rbc/tests/test_omnisci_column_list.py
```diff
@@ -71,3 +71,24 @@ def test_columns_sum(omnisci, T, variant):
     actual = list(zip(*result))[0]
 
     assert expected == actual
+
+
+@pytest.mark.xfail
+def test_columnlist_enumerate(omnisci):
```
pearu, line 77: What is the issue with ColumnList case?

rbc/tests/test_omnisci_column_basic.py
```diff
@@ -853,3 +853,21 @@ def col_dtype_fn(x, m, y):
...
+        sz = len(x)
+        set_output_row_size(sz)
+        for i, e in enumerate(x):
+            y[i] = e
+        return sz
+
+    _, result = omnisci.sql_execute(
+        f'select * from table(col_enumerate(cursor(select i4 from {omnisci.table_name})))')
+    _, expected_result = omnisci.sql_execute(
+        f'select i4 from {omnisci.table_name}')
```
pearu, line 872: The row order of `select` is undefined and to ensure correct order (that `enumerate` needs), one should use
```
select rowid, i4 from {omnisci.table_name} order by rowid;
```

pearu (commented):
I have a question and remark about ordering of rows in select.

guilhermeleobas:
Supporting `enumerate(ColumnList)` will require a bit more of work, and I'll do it in a later PR.

rbc/tests/test_omnisci_column_basic.py
pearu: Yes. Now the `expected_result` ought to be uniquely defined. The query
```
select * from table(col_enumerate(cursor(select i4 from {omnisci.table_name})))
```
is vulnerable to the same effect but it could be that table functions results are always ordered. If not, at some point we should see this test breaking, until then, let's leave it as it is.

rbc/tests/test_omnisci_column_list.py
```diff
@@ -71,3 +71,24 @@ def test_columns_sum(omnisci, T, variant):
...
+    #         for i, e in enumerate(col):
+    #             out[i] = e if j == 0 else out[i] + e
+    #         # for i in range(lst.nrows):
+    #         #     out[i] = col[i] if j == 0 else out[i] + col[i]
+    #     return lst.nrows
+
+    query = (f'select * from table(columnlist_enumerate(cursor('
+             f'select i4, i4, i4 from {omnisci.table_name}), 1))')
+    _, result = omnisci.sql_execute(query)
+    assert list(result) == [(0,), (3,), (6,), (9,), (12,)]
```
pearu, line 94: Could you store this test function in https://github.com/xnd-project/rbc/issues/442 and remove it here to keep the tests clean of dead code?

.github/workflows/rbc_test.yml
```diff
@@ -105,7 +105,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     timeout-minutes: 20
     strategy:
-      fail-fast: true
+      fail-fast: false
```
pearu, line 108: Do we want to keep it always `false`?

pearu (approved):
I have a nit to remove dead code from tests, otherwise LGTM, thanks, @guilhermeleobas !

.github/workflows/rbc_test.yml
guilhermeleobas: I think it is better for debugging when fail-fast is false.

rbc/tests/test_omnisci_column_list.py
guilhermeleobas: done

pearu (approved):
LGMT! Thanks, @guilhermeleobas !

PR #441, opened 2022-02-16
Improve IDE support with explicit parameters in signature

### LINKED ISSUE #437: Improve IDE support on user facing objects
Currently, IDE complains about a few things such as:

* string annotations when decorating a function
* possible arguments for various objects (ex: `devices`)

tupui:
Closes #437

These were the main points. Not sure if we need more for now (`Signature` could be improved but we don't really want users to handle this object).

### DIFF
--- a/rbc/remotejit.py
+++ b/rbc/remotejit.py
@@ -709,12 +709,12 @@ class RemoteJIT:
             self._targets = self.retrieve_targets()
         return self._targets
 
-    def __call__(self, *signatures, **options):
+    def __call__(self, *signatures, devices=None, local=False, templates=None):
         """Define a remote JIT function signatures and template.
 
         Parameters
         ----------
-        signatures : tuple
+        signatures : str or object
           Specify signatures of a remote JIT function, or a Python
           function as a template from which the remote JIT function
           will be compiled.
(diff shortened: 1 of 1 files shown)


rbc/remotejit.py
```diff
@@ -723,8 +723,9 @@ def __call__(self, *signatures, **options):
         ------------------
         local : bool
         devices : list
-          Specify device names for the given set of signatures.
-        templates : dict
+          Specify device names for the given set of signatures. Possible
+          values are 'cpu', 'gpu'.
+        templates : dict(str, list(str)) or list(str)
```
pearu, line 728: `list(str)` does not seem to be possible, hence:
```suggestion
        templates : dict(str, list(str))
```

rbc/remotejit.py
```diff
@@ -741,12 +742,20 @@ def __call__(self, *signatures, **options):
...
             s = Signature(self.local)
         else:
             s = Signature(self)
-        devices = options.get('devices')
-        options, templates = extract_templates(options)
+        options = {
+            "local": local,
+            "devices": devices,
+            "templates": templates.get('templates') or templates
+        }
```
pearu, line 753: Nit:
```suggestion
        options = dict(
            local=local,
            devices=devices,
            templates=templates.get('templates') or templates
        )
```

pearu (changes requested):
I have a couple of nits but otherwise LGTM.

rbc/remotejit.py
tupui: Actually it is when you do `T=['...']`. `T` is getting caught by `templates`.

tupui (commented):
Thanks @pearu!

rbc/remotejit.py
tupui: Wait no. My mistake indeed!

tupui: Ok I checked and if you do:

`ljit('T(T)', T=['i8', 'i4'])`

then `templates={T=['i8', 'i4']}`

PR #446, opened 2022-02-23
Prevent `typingctx` arg from the signature in the docs

### LINKED ISSUE #444: Remove `typingctx` from the signature of functions exported using `External`
image

guilhermeleobas:
Fixes #444 

Before:
image

After:
image


rbc/externals/__init__.py
```diff
@@ -48,4 +49,10 @@ def codegen(context, builder, sig, args):
     exec(fn_str, module_globals)
     fn = module_globals[fname]
     fn.__doc__ = doc
+
+    # https://chriswarrick.com/blog/2018/09/20/python-hackery-merging-signatures-of-two-python-functions/
+    # https://stackoverflow.com/questions/1409295/set-function-signature-in-python
```
tupui, line 54: FYI, StackOverflow's license is not compatible with BSD. So it's better to not say anything like that.
```suggestion
```

rbc/externals/__init__.py
```diff
@@ -48,4 +49,10 @@ def codegen(context, builder, sig, args):
...
+    # this is to remove `typingctx` from showing up in the docs
+    sig = inspect.signature(fn)
+    fn.__signature__ = sig.replace(parameters=tuple(sig.parameters.values())[1:])
```
tupui, line 57: Can we not look for `typingctx`? But if this function is strictly used in cases where we have `typingctx` then fine.

tupui (commented):
Thanks @guilhermeleobas. Some minor comments.

rbc/externals/__init__.py
guilhermeleobas: The function creates Numba intrinsics. Thus, all functions will have the typingctx as first argument. I can put an assertion just to make sure

tupui: ok no problem then.

tupui (approved):

rbc/externals/__init__.py
pearu: Adding a check that ensures that the dropped argument is indeed typingctx, makes sense to me, otherwise, the silent removal is a source of a-hard-to-find-bug. Since it is an internal integrity check and it would be too hard to implement checks on all internal implementations, assert is better than nothing, imho.

pearu (approved):
LGTM!

guilhermeleobas:
Failure doesn't seem to be related to this PR.

PR #447, opened 2022-03-02
Support HeavyDB

pearu:
~To test rbc against HeavyDB (former OmnisciDB) server, define `HEAVYAI_BRAND` environment variable.~
~When `HEAVYAI_BRAND` is not defined or contains a value `omnisci` then rbc expects that OmniscDB 5.10.2 or older server is running.~

rbc detects HeavyDB brand using the server version information.

### DIFF
--- a/rbc/omniscidb.py
+++ b/rbc/omniscidb.py
@@ -223,7 +223,7 @@ def get_client_config(**config):
 
     if 'dbname' not in config:
         version = get_heavydb_version(host=config['host'], port=config['port'])
-        if version[:2] >= (6, 0):
+        if version is not None and version[:2] >= (6, 0):
             if version[:3] == (6, 0, 0) and version_date(version) < 20220301:
                 # TODO: remove this if-block when heavydb 6.0 is released.
                 config['dbname'] = 'omnisci'
--- a/rbc/omniscidb.py
+++ b/rbc/omniscidb.py
@@ -355,8 +355,14 @@ service Omnisci {
         multiplexed=False,
         thrift_content=thrift_content,
         socket_timeout=60000)
-    version = client(Omnisci=dict(get_version=()))['Omnisci']['get_version']
-    _cache[host, port] = version = parse_version(version)
+    try:
+        version = client(Omnisci=dict(get_version=()))['Omnisci']['get_version']
+    except Exception as msg:
+        print(f'failed to get heavydb version[host={host}, port={port}]: {msg}')
+        version = None
+    else:
+        version = parse_version(version)
+    _cache[host, port] = version
     return version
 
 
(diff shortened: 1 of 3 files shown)


guilhermeleobas:
> export HEAVYAI_BRAND=heavyai

Could we automatically switch to heavyai when the server version is greater than `5.10.*`?

pearu:
> Could we automatically switch to heavyai when the server version is greater than `5.10.*`?

I guess we can. Since omniscidb and heavydb use different `dbname`'s, it will require trying out different configurations when connecting to the server. I'll give it a try...

rbc/omniscidb.py
```diff
@@ -162,53 +167,70 @@ def get_client_config(**config):
...
+            for k in ['debug', 'use_host_target']:
+                if k in rbc and k not in caller_config:
+                    config[k] = rbc.getboolean(k)
+
+    if 'dbname' not in config:
+        version = get_heavydb_version(host=config['host'], port=config['port'])
+        if version[:2] >= (6, 0):
+            if version[:3] == (6, 0, 0) and version_date(version) < 20220301:
+                # TODO: remove this if-block when heavydb 6.0 is released.
+                config['dbname'] = 'omnisci'
```
guilhermeleobas, line 229: Can you open an issue to track this?

rbc/omniscidb.py
```diff
@@ -314,7 +336,31 @@ def __repr__(self):
...
+'''
+    client = ThriftClient(
+        host=host,
+        port=port,
+        multiplexed=False,
+        thrift_content=thrift_content,
+        socket_timeout=60000)
+    version = client(Omnisci=dict(get_version=()))['Omnisci']['get_version']
+    _cache[host, port] = version = parse_version(version)
+    return version
```
guilhermeleobas, line 360: I guess this is needed because at this point one cannot use `heavyai.get_version()`, right?

guilhermeleobas (approved):
Good work, pearu. I wonder if it makes sense to rename the `rbc/tests/*_omnisci_*` test files and specific functions:
* `rbc.externals.omniscidb` -> `rbc.externals.heavydb`
* `rbc.omnisci_backend` -> `rbc.heavy_backend`

rbc/omniscidb.py
pearu: Correct. In fact, at the point when executing `heavyai.get_version()`, we know what is the brand of heavydb but here we only know what is `host:port` of the heavydb server.

pearu: In addition, if `dbname` would be the same for different heavydb brands, we would not need this function.

rbc/omniscidb.py
pearu: Sure: https://github.com/xnd-project/rbc/issues/450

pearu:
> I wonder if it makes sense to rename the `rbc/tests/*_omnisci_*` test files and specific functions:

Yes, see https://github.com/xnd-project/rbc/issues/451

Issue #443, opened 2022-02-18
Derive default type from fill value


guilhermeleobas:
From https://github.com/xnd-project/rbc/pull/436/#discussion_r809846919

```
should we derive the default dtype from the dtype of fill_value? numpy does this, for instance.
```

PR #452, opened 2022-03-10
Derive dtype from fill_value

guilhermeleobas:
Fix issue #443

### DIFF
--- a/rbc/irtools.py
+++ b/rbc/irtools.py
@@ -1,6 +1,7 @@
 # Author: Pearu Peterson
 # Created: February 2019
 
+import sys
 import os
 import re
 import warnings
@@ -291,7 +292,8 @@ def compile_instance(func, sig,
     # set CAPTURED_ERROS to "new_style"
     # https://github.com/numba/numba/pull/7397
     os.environ["CAPTURED_ERRORS"] = "new_style"
-    config.reload_config()
+    if 'numba' not in sys.modules:
+        config.reload_config()
 
     fname = func.__name__ + sig.mangling()
     args, return_type = sigutils.normalize_signature(
--- a/rbc/irtools.py
+++ b/rbc/irtools.py
@@ -290,7 +290,7 @@ def compile_instance(func, sig,
 
     # set CAPTURED_ERROS to "new_style"
     # https://github.com/numba/numba/pull/7397
-    os.environ["CAPTURED_ERRORS"] = "old_style"
+    os.environ["CAPTURED_ERRORS"] = "new_style"
     config.reload_config()
 
     fname = func.__name__ + sig.mangling()
--- a/rbc/irtools.py
+++ b/rbc/irtools.py
@@ -1,6 +1,7 @@
 # Author: Pearu Peterson
 # Created: February 2019
 
+import os
 import re
 import warnings
 from contextlib import contextmanager
@@ -14,7 +15,7 @@ from . import libfuncs
 from rbc import externals
 from numba.core import codegen, cpu, compiler_lock, \
     registry, typing, compiler, sigutils, cgutils, \
-    extending, imputils
+    extending, imputils, config
 from numba.core import errors as nb_errors
 
 
@@ -286,6 +287,11 @@ def compile_instance(func, sig,
         flags.set('no_compile')
         flags.set('no_cpython_wrapper')
         flags.set('no_cfunc_wrapper')
+
+    # set CAPTURED_ERROS to "new_style"
+    # https://github.com/numba/numba/pull/7397
+    os.environ["CAPTURED_ERRORS"] = "old_style"
+    config.reload_config()
 
     fname = func.__name__ + sig.mangling()
     args, return_type = sigutils.normalize_signature(
--- a/rbc/irtools.py
+++ b/rbc/irtools.py
@@ -1,8 +1,6 @@
 # Author: Pearu Peterson
 # Created: February 2019
 
-import sys
-import os
 import re
 import warnings
 from contextlib import contextmanager
@@ -16,7 +14,7 @@ from . import libfuncs
 from rbc import externals
 from numba.core import codegen, cpu, compiler_lock, \
     registry, typing, compiler, sigutils, cgutils, \
-    extending, imputils, config
+    extending, imputils
 from numba.core import errors as nb_errors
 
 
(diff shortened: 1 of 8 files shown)


tupui (approved):
LGTM, it's a good addition and the tests seem comprehensive.

rbc/typesystem.py
```diff
@@ -1109,6 +1109,8 @@ def fromnumba(cls, t):
         if isinstance(t, nb.types.misc.RawPointer):
             if t == nb.types.voidptr:
                 return cls(cls(), '*')
+        if isinstance(t, nb.types.StringLiteral):
+            return cls.fromstring(t.literal_value)
```
pearu, line 1113: This does not make sense:
```
>>> import numba
>>> import rbc.typesystem as ts
>>> ti=ts.TargetInfo('cpu')
>>> s=numba.types.StringLiteral('abc')
>>> s
Literal[str](abc)
>>> with ti: ts.Type.fromnumba(s)
... 
Type('abc',)
```
that is, a numba string literal becomes a type that name is the value of the string literal.

The `fromnumba` method should be applied to numba types only. For constructing Type from a value, one should use `Type.fromvalue` method.

So:
```suggestion
```

rbc/irtools.py
```diff
@@ -287,6 +288,11 @@ def compile_instance(func, sig,
         flags.set('no_cpython_wrapper')
         flags.set('no_cfunc_wrapper')
 
+    # set CAPTURED_ERROS to "new_style"
+    # https://github.com/numba/numba/pull/7397
+    os.environ["CAPTURED_ERRORS"] = "new_style"
```
pearu, line 293: Currently, the environment is set and the config is reloaded on all `compile_instance` calls.

I suggest setting the environment before importing numba, that is, in `rbc/__init__.py` have the following code:
```python
os.environ["CAPTURED_ERRORS"] = "new_style"
if 'numba' is sys.modules:
    numba.core.config.reload_config()
```
(untested, just illustrates the idea).

pearu (changes requested):
Thanks, @guilhermeleobas . I have a couple of suggestions/nits but overall looks good.

rbc/stdlib/creation_functions.py
```diff
@@ -19,6 +20,21 @@
...
+        # determine from fill_value
+        if fill_value is None:
+            # return the default floating-point data type
+            return types.double
+        else:
+            return asnumbatype.typeof(fill_value)
+    else:
+        if isinstance(dtype, types.UnicodeType):
+            raise errors.RequireLiteralValue(dtype)
+        else:
```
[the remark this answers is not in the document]
guilhermeleobas: This is for the case one calls a creation function and provide a dtype value (i.e. `array_api.zeros(3, dtype='int32')`). When Numba is compiling the code, it will first type the argument as Unicode and if it fails, with a literal value. Since we want the second case to succeed. Hence, the exception is needed.

rbc/tests/test_omnisci_array_functions.py
```diff
@@ -161,14 +163,95 @@ def test_array_methods(omnisci, method, signature, args, expected):
...
+             'array')
+
+
+@pytest.mark.parametrize('dtype', ('int32_t', 'float'))  # test all dtypes?
+@pytest.mark.parametrize('func_name', functions)
+def test_literal_string_dtype(omnisci, dtype, func_name):
+    omnisci.reset()
+
+    if func_name == 'array':
+        return
```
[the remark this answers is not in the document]
guilhermeleobas: I've removed this function from the source code for the following reasons:
1) It's not part of the Array API standard
2) No test coverage
3) The current implementation (copies one array into other) is too simple to match what `numpy.array` does
https://numpy.org/doc/stable/reference/generated/numpy.array.html

Let me know what are your thoughts about this.

rbc/irtools.py
```diff
@@ -287,6 +289,12 @@ def compile_instance(func, sig,
...
+    if 'numba' not in sys.modules:
```
pearu, line 295: This does not make sense. When the code execution hits this, numba is surely imported.

The `reload_config` call should happen at the early stage of loading `rbc` modules, say, in `rbc/__init__.py` (so that we do it only once) and only if `numba` has been imported before `rbc` and after resetting CAPTURED_ERRORS env variable (otherwise there is nothing new to reload).

In addition, when a user specifies `CAPTURED_ERRORS` env variable, we should respect that:
```python
if "CAPTURED_ERRORS" not in os.environ:
    os.environ["CAPTURED_ERRORS"] = "new_style"
```

rbc/stdlib/creation_functions.py
```diff
@@ -4,21 +4,38 @@
...
+        if fill_value is None:
+            # return the default floating-point data type
+            return types.double
+        else:
+            return asnumbatype.typeof(fill_value)
+    else:
+        if isinstance(dtype, types.UnicodeType):
+            raise errors.RequireLiteralValue(dtype)
+        elif isinstance(dtype, types.StringLiteral):
+            return typesystem.Type.fromstring(dtype.literal_value).tonumba()
```
pearu, line 34: Yes, this makes sense.

rbc/stdlib/creation_functions.py
```diff
@@ -4,21 +4,38 @@
...
+        else:
+            return typesystem.Type.fromobject(dtype).tonumba()
```
pearu, line 36: Here `dtype` can be any object, right? For instance, a user could call `array_api.zeros(3, dtype=123)` and it will fail at calling `typesystem.Type.fromobject`.

I suggest that withing `_determine_dtype` we explicitly check the correctness of `dtype` argument type: it can be either `str` object or `type` object, all other types (ints, floats, lists, etc) should raise a TypeError (that, if possible, would circumvent the verbose numba exception handling).

pearu (changes requested):
CAPTURED_ERRORS logic is wrong, otherwise looks good!

rbc/stdlib/creation_functions.py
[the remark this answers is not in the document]
pearu: Thanks for the explanation!

rbc/irtools.py
guilhermeleobas: Would it be ok if we keep the `config.reload_config()` here? It's a cheap operation that will not slow down RBC.

pearu: I could see it being cheap but reading this code can be confusing: each call to a function sets environment variable and reloads config, and one may wonder if this is intentional and why.

I would be ok with:
```
if "CAPTURED_ERRORS" not in os.environ:
    os.environ["CAPTURED_ERRORS"] = "new_style"
    config.reload_config()
```
that ensures that the env set and reload_config is called exactly once (unless users explicitly do `del os.environ["CAPTURED_ERRORS"]`, for instance) plus if you would be ok if at some point this code is moved elsewhere :)

pearu (approved):

PR #453, opened 2022-03-15
Add unimplemented functions from array_api

guilhermeleobas:
This doesn't implement all the functions, only include the files so that they can show up in the docs.

### DIFF
--- a/rbc/stdlib/array_attributes.py
+++ b/rbc/stdlib/array_attributes.py
@@ -1,0 +1,5 @@
+"""
+https://data-apis.org/array-api/latest/API_specification/array_object.html#attributes
+"""
+
+__all__ = []
--- a/rbc/stdlib/array_methods.py
+++ b/rbc/stdlib/array_methods.py
@@ -1,0 +1,5 @@
+"""
+https://data-apis.org/array-api/latest/API_specification/data_type_functions.html
+"""
+
+__all__ = []
--- a/rbc/stdlib/array_api.py
+++ b/rbc/stdlib/array_api.py
@@ -2,10 +2,17 @@
 Array API for rbc.
 """
 
-from .datatypes import *  # noqa: F401, F403
 from .constants import *  # noqa: F401, F403
 from .creation_functions import *  # noqa: F401, F403
+from .datatypes import *  # noqa: F401, F403
+from .data_type_functions import *  # noqa: F401, F403
 from .elementwise_functions import *  # noqa: F401, F403
+from .linear_algebra_functions import *  # noqa: F401, F403
+from .manipulation_functions import *  # noqa: F401, F403
+from .searching_functions import *  # noqa: F401, F403
+from .set_functions import *  # noqa: F401, F403
+from .sorting_functions import *  # noqa: F401, F403
 from .statistical_functions import *  # noqa: F401, F403
+from .utility_functions import *  # noqa: F401, F403
 
-__all__ = [s for s in dir() if not s.startswith('_')]
+__all__ = [s for s in dir() if not s.startswith("_")]
--- a/rbc/stdlib/data_type_functions.py
+++ b/rbc/stdlib/data_type_functions.py
@@ -1,0 +1,72 @@
+"""
+https://data-apis.org/array-api/latest/API_specification/data_type_functions.html
+"""
+from rbc.stdlib import Expose
+
+__all__ = [
+    "astype",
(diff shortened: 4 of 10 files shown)


rbc/stdlib/array_methods.py
```diff
@@ -0,0 +1,7 @@
+"""
+Array API specification for array methods.
+
+https://data-apis.org/array-api/latest/API_specification/data_type_functions.html
+"""
+
+__all__ = []
```
tupui, line 7: Why is this file needed? Isn't it already `rbc/stdlib/data_type_functions.py`?

rbc/stdlib/array_attributes.py
```diff
@@ -0,0 +1,7 @@
+"""
+Array API specification for array attributes.
+
+https://data-apis.org/array-api/latest/API_specification/array_object.html#attributes
+"""
+
+__all__ = []
```
tupui, line 7: Same. This is supposed to be `rbc/stdlib/data_type_functions.py` no?

tupui (commented):
Ok so the goal is to have the full Array API defined but raise if something is not implemented. Sounds good.

rbc/stdlib/array_attributes.py
guilhermeleobas: Same. https://data-apis.org/array-api/latest/API_specification/array_object.html#attributes

rbc/stdlib/array_methods.py
guilhermeleobas: Sort of. I still need to include the code for unimplemented methods and attributes.

tupui: ok thanks for confirming, I was not sure as the PR was not in draft.

guilhermeleobas: Sorry about that. In the end I remove this file and moved the attributes to `rbc/omnisci_backend/omnisci_array.py`

tupui (approved):
LGTM

rbc/omnisci_backend/omnisci_array.py
```diff
@@ -31,7 +35,101 @@ def buffer_extra_members(self):
...
+
+    .. code-block:: C
+
+        {
+            T* data,
+            int64_t size,
+            int8_t is_null
+        }
+
+    Array is a contiguous block of memory that behaves like a python list.
```
pearu, line 49: Python list may be a little too high-level object to make this relation. For instance, python lists can be concatenated while contiguous blocks of memories can't, etc. How about this: "Array holds a contiguous block of memory and it implements a subset of the array protocol."

rbc/omnisci_backend/omnisci_array.py
```diff
@@ -31,7 +35,101 @@ def buffer_extra_members(self):
...
+    Example
+
+
+    .. code-block:: python
+
+        from numba import types
+        from rbc.omnisci_backend import Array
+
+        @omnisci('int64[](int64)')
+        def create_and_fill_array(size):
```
pearu, line 60:
```suggestion
        def my_arange(size):
```

rbc/omnisci_backend/omnisci_array.py
```diff
@@ -31,7 +35,101 @@ def buffer_extra_members(self):
...
+        Hardware device the array data resides on.
+        """
+        pass
+
+    @property
+    def mT(self):
+        """
+        ❌ Not implemented
+
+        Transpose of a amtrix (or a stack of matrices).
```
pearu, line 94:
```suggestion
        Transpose of a matrix (or a stack of matrices).
```
What is a matrix in omnisci context? Array of arrays? In any case, since we don't have matrices yet, `mT` should be marked as "Not supported" and mentioning matrix in a doc string does not make sense yet.

pearu (changes requested):
I have only a few comments, otherwise looks good.

While implementing and documenting Array API in the HeavyDB context, I think we should clearly distinguish the "Not implemented" and "Not supported" cases. The "not implemented" means that the corresponding feature can be implemented if taken some time for it, while "not supported" means that the corresponding feature does not makes sense because the underlying datastructures do not support it. For instance, OmnisciDB currently is able to support only 1-D arrays, and hence, we should not give hope to users that inherently N-D operations such as matrix operations can be supported via raising an rbc issue.

pearu (approved):

PR #458, opened 2022-03-17
Remove AutoFreeBuffers pass

guilhermeleobas:
This PR removes the broken `AutoFreeBuffers` pass and introduce a metadata for heavydb to release memory after the UD[T]F is executed.

Requires https://github.com/heavyai/heavydb-internal/pull/6424


pearu (changes requested):
I have a couple of code quality suggestions.

PR #459, opened 2022-03-18
Remove truncate instruction - Fix issue #457

### LINKED ISSUE #457: RBC is truncating values on Buffer.__setitem__ if the rhs type is bigger than the container bit width
I just notice the fix I wrote for #197 is truncating elements if the buffer bit width dtype is smaller than element bit width. For instance, in the code below, RBC will emit a `trunc` instruction to truncate the rhs value.

```python
@omnisci('int8[](int32)')
def arange(sz):
    arr = array_api.zeros(sz, dtype='int8')
    for i in range(sz):
        # assignment is invalid
        arr[i] = 123456  # the rhs is typed as int64 and doesn't fit on an int8.
    return arr
```

The culprit of it is the function `truncate_or_extend` that is truncating elements at will. This is wrong, and RBC shouldn't be doing this at all!

https://github.com/xnd-project/rbc/blob/c84a11a376d7e4d165743521bf67f752cb697fc4/rbc/omnisci_backend/omnisci_buffer.py#L428-L449

[JAX](https://jax.readthedocs.io/en/latest/design_notes/type_promotion.html) has a good page on type promotion which RBC can base of.

guilhermeleobas:
As title. Closes #457

### DIFF
--- a/rbc/heavyai/buffer.py
+++ b/rbc/heavyai/buffer.py
@@ -33,6 +33,7 @@ from rbc.targetinfo import TargetInfo
 from llvmlite import ir as llvm_ir
 from numba.core import datamodel, cgutils, extending, types, imputils
 from rbc.errors import UnsupportedError
+import warnings
 
 int8_t = ir.IntType(8)
 int32_t = ir.IntType(32)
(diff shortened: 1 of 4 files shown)


rbc/omnisci_backend/omnisci_bytes.py
```diff
@@ -33,14 +40,50 @@ def match(self, other):
...
+            ptr_t,
+            size_t,
+            *extra_members
+        )
+        buffer_type._params['NumbaType'] = BufferType
+        buffer_type._params['NumbaPointerType'] = BytesPointer
+        numba_type = buffer_type.tonumba(bool_is_int8=True)
+        if self.pass_by_value:
+            return numba_type
+        return BytesPointer(numba_type)
```
pearu, line 57:
```suggestion
```

Instead of copying the whole method, I'd suggest extending `OmnisciBufferType.tonumba` as follows:
```python
class OmnisciBufferType:
    @property
    def numba_pointer_type(self):
        return BufferPointer
    ...
    def tonumba(self):
        ...
       buffer_type._params['NumbaPointerType'] = self.numba_pointer_type
       ...

class OmnisciBytesType(OmnisciBufferType):
    ...
    @property
    def numba_pointer_type(self):
        return BytesPointer
    ...
```

rbc/tests/test_omnisci_array.py
```diff
@@ -404,13 +405,18 @@ def fn_issue197(x):
 
     omnisci(f'{typ}[]({typ}[])')(fn_issue197)
 
-    _, result = omnisci.sql_execute(
-        f'SELECT {col}, {fn_name}({col}) FROM {omnisci.table_name};'
-    )
+    if 'trunc' in suffix:
+        with pytest.raises(TypingError, match='mismatching types'):
+            omnisci.register()
```
pearu, line 410: What is going on here? Does this indicate that something is not cleaned up properly from the previous tests?

pearu (changes requested):
I have a suggestion to improve tonumba method of OmnisciBufferType and a question.

guilhermeleobas:
From today's meeting, it was decided to leave the code as it is. Numba has the same behavior.

PR #464, opened 2022-03-24
Introduce RemoteHeavyDB.reconnect. Test UDF registration by non-admin users.

pearu:
As in the title.

### DIFF
--- a/rbc/heavydb.py
+++ b/rbc/heavydb.py
@@ -442,6 +442,21 @@ class RemoteHeavyDB(RemoteJIT):
         # An user-defined device-LLVM IR mapping.
         self.user_defined_llvm_ir = {}
 
+    def __repr__(self):
+        return (f'{type(self).__name__}(user={self.user!r}, password="{"*"*len(self.password)}",'
+                f' host={self.host!r}, port={self.port}, dbname={self.dbname!r})')
+
+    def reconnect(self, user=None, password=None, dbname=None):
+        """Return another RemoteHeavyDB instance with possibly different
+        connection credentials.
+        """
+        return type(self)(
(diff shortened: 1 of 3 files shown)


rbc/heavydb.py
```diff
@@ -442,6 +442,21 @@ def __init__(self,
...
         self.user_defined_llvm_ir = {}
 
+    def __repr__(self):
+        return (f'{type(self).__name__}(user={self.user!r}, password="{"*"*len(self.password)}",'
+                f' host={self.host!r}, port={self.port}, dbname={self.dbname!r})')
+
+    def reconnect(self, user=None, password=None, dbname=None):
+        """Return another RemoteHeavyDB instance with possibly different
+        connection credentials.
+        """
```
guilhermeleobas, line 452: Perhaps raise an exception if the server version is < (5, 9) here.

guilhermeleobas (approved):
Looks good to me. I only have one comment regards raising an exception if the server version doesn't have this feature.

rbc/heavydb.py
pearu: No need, the reconnect works fine with any server (see test_reconnect).

The test_non_admin_user requires 5.9+ because SQL "DROP USER.." statement did not work with older versions than 5.9.

tupui (approved):
LGTM, thanks Pearu.

I was wondering why we needed to use the same object to create a new connection, but it can be handy to keep the credentials from a user perspective.

To be clear, we don't need this for non admin users to register a UDF right?

pearu:
> To be clear, we don't need this for non admin users to register a UDF right?

Correct. The reconnect feature is mostly useful for testing or for some more complex setups.

PR #465, opened 2022-03-25
Introduce RBC_DEBUG env var

guilhermeleobas:
As title

### DIFF
--- a/doc/envvars.rst
+++ b/doc/envvars.rst
@@ -1,0 +1,11 @@
+
+Environment variables
+=====================
+
+Debugging
+---------
+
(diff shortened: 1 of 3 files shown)


pearu (approved):
LGTM!

Here is a list of other environment variables that we should document:
- HEAVYDB_CLIENT_CONF - see rbc/heavydb.py:get_client_config
- RBC_TESTS_FULL - see rbc/tests/heavyai/test_heavyai.py
- OMNISCIDB_DEV_LABEL - see rbc/tests/__init__.py (requires rebranding)

UPDATE: created https://github.com/xnd-project/rbc/issues/466

Issue #467, opened 2022-03-28
Rename heavyai to heavydb


guilhermeleobas:
Currently, the RBC package contains a `heavyai` subpackage and a `heavydb.py` file. I propose to rename the subpackage to `heavydb` and keep everything there.

PR #470, opened 2022-03-29
Fix issue #467 - Move imports to rbc.heavydb

guilhermeleobas:
As title.

* Moves all imports from `rbc.heavai.*` to `rbc.heavydb.*`
* Rename the test folder `rbc/tests/heavyai` to `rbc/tests/heavydb`
* Includes a test file `rbc/tests/heavydb/test_imports.py` that ensures imports for `Array`, `ArrayPointer`, `Bytes`, `Column` are working as expected. Some of those imports are used in UD[T]Fs to create objects at runtime, others are used internally in jitted functions.


rbc/tests/heavydb/test_array_methods.py
```diff
@@ -1,7 +1,8 @@
 import pytest
 import numpy as np
 from rbc.tests import heavydb_fixture
-from rbc.heavyai import Array
+from rbc.heavydb import Array
+from rbc.stdlib import array_api  # noqa: F401
```
tupui, line 5: Why is this needed?

rbc/tests/heavydb/test_imports.py
```diff
@@ -0,0 +1,17 @@
...
+    'ArrayPointer',
+    'Bytes',
+    'Column',
+]
+
+
+@pytest.mark.parametrize('cls', classes)
+def test_heavydb_imports(cls):
+    try:
+        exec(f'from rbc.heavydb import {cls}')
```
tupui, line 15: I think `importlib` should be used instead.

tupui (commented):
I went over all the files, LGTM. Just 2 remarks.

Ultimately, why don't we just move `heavydb` into `heavydb_backend`?

rbc/tests/heavydb/test_array_methods.py
guilhermeleobas: This initializes the array methods implementations, otherwise the tests will fail.

rbc/tests/heavydb/test_imports.py
guilhermeleobas: Done!

tupui (approved):
LGTM now. Although I see you adding files again. Are you missing more?

pearu (approved):
LGTM!

Re @tupui comment: Ultimately, why don't we just move heavydb into heavydb_backend?

Yes, I think it makes sense but as a follow-up PR:
- move `heavydb.py` to `heavydb_backend/remote_heavydb.py` (or suggest a better name)
- rename `heavydb_backend/` to `heavydb/`
- expose the UI of `heavydb.py` via `heavydb/__init__.py`

guilhermeleobas:
> LGTM now. Although I see you adding files again. Are you missing more?

I've messed up during rebase 😆

Issue #473, opened 2022-03-29
Do not override user-specified arguments in RemoteHeavyDB


guilhermeleobas:
In general, we should not override user-specified arguments (while unlikely, a user of omniscidb v 5.9 may have explicitly specified `dbname=heavyai` and the suggested behavior is a bug).

Here is an approach to resolve the problem, in general:
```
class default_str(str):
    pass

def default(obj):
    if isinstance(obj, str):
        return default_str(obj)
    else:
        raise NotImplementedError(str(type(obj)))

class RemoteHeavyDB(RemoteJIT):
    ...
    def __init__(self, ..., dbname=default('heavyai')):
        ...
        if _version and _version < (6, 0, 0) and isinstance(dbname, default_str):
            dbname = 'omnisci'
        ...
```

_Originally posted by @pearu in https://github.com/xnd-project/rbc/pull/463#discussion_r835954546_

PR #474, opened 2022-03-29
Do not override user specified `dbname`

guilhermeleobas:
Fix issue #473

### DIFF
--- a/rbc/heavydb.py
+++ b/rbc/heavydb.py
@@ -414,15 +414,18 @@ class RemoteHeavyDB(RemoteJIT):
                  password='HyperInteractive',
                  host='localhost',
                  port=6274,
-                 dbname='heavyai',
+                 dbname=None,  # defaults to 'heavyai'
                  **options):
         self.user = user
         self.password = password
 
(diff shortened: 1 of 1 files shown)


pearu (changes requested):
I think this approach is better than before but it basically postpones tackling the issue to the future: at some point, we'll need to update notebooks again. Once heavydb 6.0 is released, we'll need to change dbname back.

The underlying problem is that notebooks are expected to use the default dbname and the default dbname depends on heavydb version. And ideally, the notebooks ought to be functional for all omnscidb/heavydb versions.

Here's a suggestion for a more permanent fix:
```python
class RemoteHeavyDB(RemoteJIT):
    ...
    def __init__(self,
                 user='admin',
                 password='HyperInteractive',
                 host='localhost',
                 port=6274,
                 dbname=None,  # defaults to 'heavyai'
                 **options):
        if dbname is None:
            v = get_heavydb_version()
            if v and v[0] < 6:
                dbname = 'omnisci'
            else:
                dbname = 'heavyai'
```

pearu (approved):
LGTM! (let's wait for CI to become green)

PR #478, opened 2022-03-31
Restore remotejit section 

guilhermeleobas:
This PR restores the remotejit section with a single machine just to make sure no errors happen when heavydb server is not available.


pearu (approved):

PR #481, opened 2022-03-31
Remove pytest.mark.skip from test

guilhermeleobas:
This restores a set of tests that were previously marked to be skipped


tupui (approved):
Tests are green so good thing then to remove 👍

PR #483, opened 2022-04-22
cast boolean literal to Boolean8

guilhermeleobas:
As title.


pearu (approved):
LGTM!

Any idea why one CI job was canceled?

guilhermeleobas:
That's odd. For me it is still running.

PR #485, opened 2022-04-22
Add thrift endpoints to retrieve registered udfs/udtfs

guilhermeleobas:
As title.

### DIFF
--- a/rbc/heavydb/remoteheavydb.py
+++ b/rbc/heavydb/remoteheavydb.py
@@ -505,20 +505,32 @@ class RemoteHeavyDB(RemoteJIT):
         return self._thrift_client
 
     def get_table_function_names(self, *, runtime_only=False):
+        """Return a list of table function names registered on the server.
+        If `runtime_only` is True, then, only runtime UDTF names are returned
+        """
         if runtime_only:
             return self.thrift_call('get_runtime_table_function_names', self.session_id)
         else:
             return self.thrift_call('get_table_function_names', self.session_id)
 
     def get_table_function_details(self, *args):
+        """Given a list of table function names, returns a
+        TUserDefinedTableFunction for each UDTF name
+        """
         endpoint = 'get_table_function_details'
         return self.thrift_call(endpoint, self.session_id, list(args))
 
     def get_udf_names(self, *, runtime_only=False):
+        """Return a list of user defined function names registered on the server.
+        If `runtime_only=True`, then, only runtime UDF names are returned.
+        """
         endpoint = 'get_runtime_udf_names' if runtime_only else 'get_udf_names'
         return self.thrift_call(endpoint, self.session_id)
 
     def get_udf_details(self, *args):
+        """Given a list of user defined function names, return a
+        TUserDefinedFunction for each UDF name
+        """
         endpoint = 'get_udf_details'
         return self.thrift_call(endpoint, self.session_id, list(args))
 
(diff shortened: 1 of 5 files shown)


rbc/heavydb/remoteheavydb.py
```diff
@@ -504,6 +504,36 @@ def thrift_client(self):
             self._thrift_client = self.client
         return self._thrift_client
 
+    def get_table_function_names(self, *, runtime_only=False):
```
tupui, line 507: Personally not a fan of using `get/set` in function names. Does not look Pythonic.

rbc/heavydb/remoteheavydb.py
```diff
@@ -504,6 +504,36 @@ def thrift_client(self):
...
+            return self.thrift_call('get_table_function_names', self.session_id)
+
+    def get_table_function_details(self, *args):
+        """Given a list of table function names, returns a
+        TUserDefinedTableFunction for each UDTF name
+        """
+        endpoint = 'get_table_function_details'
+        return self.thrift_call(endpoint, self.session_id, list(args))
+
+    def get_udf_names(self, *, runtime_only=False):
```
tupui, line 523: Would there be a way to have a single function instead of one for UDF and one for UDTF? Naive approach is to call both and check for raises.

tupui (commented):
Thanks @guilhermeleobas! Small suggestions

rbc/heavydb/remoteheavydb.py
guilhermeleobas: Do you have a better naming suggestions? I was using the same names as the thrift endpoint.

tupui: I would just remove the prefix `get_`

rbc/heavydb/remoteheavydb.py
```diff
@@ -504,6 +504,36 @@ def thrift_client(self):
...
+    def table_function_names(self, *, runtime_only=False):
+        """Return a list of table function names registered on the server.
+        If `runtime_only` is True, then, only runtime UDTF names are returned
+        """
+        if runtime_only:
+            return self.thrift_call('get_runtime_table_function_names', self.session_id)
+        else:
+            return self.thrift_call('get_table_function_names', self.session_id)
+
+    def table_function_details(self, *args):
```
pearu, line 516: Nit:
```suggestion
    def table_function_details(self, *names):
```

rbc/heavydb/remoteheavydb.py
```diff
@@ -504,6 +504,36 @@ def thrift_client(self):
...
+        """Given a list of table function names, returns a
+        TUserDefinedTableFunction for each UDTF name
+        """
+        endpoint = 'get_table_function_details'
+        return self.thrift_call(endpoint, self.session_id, list(args))
```
pearu, line 521:
```suggestion
        return self.thrift_call(endpoint, self.session_id, list(names))
```

rbc/tests/heavydb/test_heavydb_thrift.py
```diff
@@ -0,0 +1,83 @@
...
+
+def test_get_udtf_details(heavydb):
+    heavydb.unregister()
+
+    @heavydb('int32(Column<int>, OutputColumn<int>)')
+    def my_udtf(x, y):
+        return 0
+
+    heavydb.register()
+    details = heavydb.table_function_details('my_udtf')
```
pearu, line 41:
```suggestion
    details = heavydb.table_function_details('my_udtf')
    assert len(details) == 1
```

pearu (changes requested):
Looks good! I have only a couple of minor nits.

pearu:
Re CI failures: try to rebase against master?

guilhermeleobas:
Current branch is already up to date with upstream/main:
```
$ git rebase upstream/main
Current branch guilhermeleobas/thrift_endpoints is up to date.
```

PR #493, opened 2022-05-29
Tests for Column of Arrays (heavydb-internal PR 6571).

pearu:
As in the title plus various updates.


guilhermeleobas:
It looks like that `test_array_copier` test is broken on HeavyDB master.

pearu:
Part of this PR is included in https://github.com/xnd-project/rbc/pull/493

PR #494, opened 2022-06-03
Add HeavyDBTimestampType

tupui:


rbc/tests/heavydb/test_timestamp.py
```diff
@@ -0,0 +1,54 @@
...
+        return Timestamp(t1 + t2)
+
+    heavydb.register()
+    # print(foo.describe())
+    # table = 'time_test'
+    # query = f'select * from table(timestamp_extract_year(cursor(select t1 from {table})));'
+    # _, result = heavydb.sql_execute(query)
+    # print(list(result))
+
+
```
guilhermeleobas, line 54: Can you rewrite this test to keep only the `timestamp_extract` and the query?

guilhermeleobas (changes requested):
Hi Pamphile, can you fix the flake8 failure and rewrite the test?

guilhermeleobas:
Can you rebase with RBC main branch? Also, restrict the test to heavydb 6.0 or greater.

tupui:
I added the skip. For the rebase, this branch is already up to date with `main`

guilhermeleobas:
@tupui, to fix the current error. Change `buffer.py::truncate_cast_or_extend` calls to

```python
if isinstance(value.type, (ir.IntType, ir.FloatType, ir.DoubleType)):
    value = truncate_cast_or_extend(builder, value, buf.type.pointee, signed)
```

guilhermeleobas:
@tupui, to fix the current failure, apply the following patch

```diff
diff --git a/rbc/heavydb/timestamp.py b/rbc/heavydb/timestamp.py
index f1764aa..62f6224 100644
--- a/rbc/heavydb/timestamp.py
+++ b/rbc/heavydb/timestamp.py
@@ -32,6 +32,9 @@ class HeavyDBTimestampType(typesystem.Type):
         return 'Timestamp'
 
 
+extending.make_attribute_wrapper(TimestampNumbaType, 'time', 'time')
+
+
 @extending.type_callable(Timestamp)
 def type_heavydb_timestamp(context):
     def typer(arg):
diff --git a/rbc/tests/heavydb/test_timestamp.py b/rbc/tests/heavydb/test_timestamp.py
index 6792591..37cbad2 100644
--- a/rbc/tests/heavydb/test_timestamp.py
+++ b/rbc/tests/heavydb/test_timestamp.py
@@ -33,7 +33,7 @@ def test_timestamp(heavydb):
     def timestamp_extract(x, y):
         set_output_row_size(len(x))
         for i in range(len(x)):
-            y[i] = x[i]
+            y[i] = x[i].time
         return len(x)
 
     heavydb.register()
@@ -41,4 +41,4 @@ def test_timestamp(heavydb):
     table = 'time_test'
     query = f'select * from table(timestamp_extract(cursor(select t1 from {table})));'
     _, result = heavydb.sql_execute(query)
-    assert list(result) == [1, 1654268400000]
+    assert list(result) == [(1,), (1654268400000000000,)]
```

tupui:
@guilhermeleobas can you have another look?

setup.py
```diff
@@ -42,7 +42,7 @@ def setup_package():
         # once the numba and llvmlite have not matched.
         install_requires = open('requirements.txt', 'r').read().splitlines()
         setup_requires = ['pytest-runner']
-        tests_require = ['pytest']
+        tests_require = ['pytest', 'packaging']
```
guilhermeleobas, line 45: Any reason to add packaging as a required dependency?

guilhermeleobas (approved):
Thanks for the work, Pamphile. I have only one commenting regarding the `packaging` package.

rbc/tests/heavydb/test_timestamp.py
```diff
@@ -0,0 +1,48 @@
...
+    if curr_version < Version("6"):
+        pytest.skip(
+            f"Requires HeavyDB >= 6, got {available_version}",
+            allow_module_level=True
+        )
+
+
+@pytest.fixture(scope='module')
+def heavydb():
+    for o in heavydb_fixture(globals(), debug=not True, load_columnar=True):
```
guilhermeleobas, line 24: `heavydb_fixture` accepts an argument `minimal_version` which you can use to skip the whole file if the server version is not what you expect.

```suggestion
@pytest.fixture(scope='module')
def heavydb():
    for o in heavydb_fixture(globals(), debug=not True, load_columnar=True, minimal_version=(6, 0)):
```

And inside `rbc/tests/heavydb/__init__.py::heavydb_fixture`, there's a function called `require_version` which checks if the given version is older than the available one. Move what you wrote here to there.

guilhermeleobas (changes requested):

setup.py
tupui: Indeeeed.

rbc/tests/heavydb/test_timestamp.py
tupui: Aoutch seeing this... I will move this in for now, but this whole thing deserves a rewrite at some point.

tupui:
I made the changes @guilhermeleobas

guilhermeleobas (approved):
Thanks Pamphile. With the introduction of `packaging`, perhaps one should revisit all the test files and fix the version comparison to use the new library.

rbc/tests/heavydb/test_timestamp.py
guilhermeleobas: You mean the fixture? I've changed it a bit on #496

tupui:
Yes that would be a good thing to do IMO. Thanks for the review 😃

@pearu do you want to have a look?

PR #496, opened 2022-06-03
Add methods to timestamp class

guilhermeleobas:
Requires #494 

No ColumnListTimestamp at the moment as heavydb doesn't support it. (cc @brenocfg)


tupui (approved):
This LGTM and I resolved the merge conflicts. I just relaunched failed jobs as it seems intermittent.

rbc/heavydb/timestamp.py
```diff
@@ -1,17 +1,258 @@
 """HeavyDB Timestamp type that corresponds to HeavyDB type timestamp."""
```
brenocfg, line 1: Very minor but shouldn't this read `RBC Timestamp type that corresponds...`? As it is it's basically a tautology :p

brenocfg (approved):
Only suggested a small change in a comment but LGTM as well

rbc/heavydb/timestamp.py
guilhermeleobas: Thanks!

PR #497, opened 2022-06-07
Update dependencies

### LINKED ISSUE #334: Remove clang and clangxx pinning on environment.yml
I've pinned clang and clangxx to 11.1 on `.conda/environment.yml` on PR #333. The reason for that is that Numba can't process LLVM 12 intermediate representation.

Once https://github.com/numba/llvmlite/issues/688 gets solved, we can remove the pinning

tupui:
General maintenance:

* Remove `six` dependency
* Update the recipe (or can we just remove the folder?)
* Remove `clang` since now we have a pure Python package
* Address deprecation of heavydb flag (missing `s`)

Closes #334


.conda/meta.yaml
```diff
@@ -1,43 +1,35 @@
...
   run:
+    - cffi
     - python
     - thriftpy2
     - tblib
-    - llvmlite
-    - numba
-    - six
+    - llvmlite >=0.29
+    - numba >=0.52
```
guilhermeleobas, line 31: RBC doesn't support numba < 0.54
```suggestion
    - numba >=0.54
```

.conda/meta.yaml
```diff
@@ -1,43 +1,35 @@
...
+    - pip
   run:
+    - cffi
     - python
     - thriftpy2
     - tblib
-    - llvmlite
-    - numba
-    - six
+    - llvmlite >=0.29
```
guilhermeleobas, line 30: I don't think we need to pin llvmlite because Numba already selects the correct version.

.conda/meta.yaml
```diff
@@ -1,43 +1,35 @@
...
-    - pytest
-    - thriftpy2
-    - tblib
-    - llvmlite >=0.29
-    - numba
-    - netifaces
+    - setuptools
+    - pip
   run:
+    - cffi
```
guilhermeleobas, line 26: Do we need `cffi`?

.conda/meta.yaml
```diff
@@ -1,43 +1,35 @@
...
   imports:
     - rbc
     - rbc.thrift
 
 requirements:
-  build:
-    - python
-    - setuptools
   host:
+    - cffi
```
guilhermeleobas, line 21: Same here. Do we need `cffi`?

.heavyai/heavyai.conf
```diff
@@ -1,4 +1,4 @@
-enable-runtime-udf = true
+enable-runtime-udfs = true
 enable-table-functions = true
 enable-dev-table-functions = true
 cpu-only = true
```
guilhermeleobas, line 4: I don't think we use this file. So, I'm in favor of removing it.

.conda/meta.yaml
```diff
@@ -1,43 +1,35 @@
 package:
   name: rbc
-  version: 0.1.0.dev2
+  version: 0.8.0
```
guilhermeleobas, line 3: I don't think this file is used anywhere, the feedstock has its own meta file. Perhaps we should just remove it.

guilhermeleobas (changes requested):
Hi Pamphile, thanks for working on this. I've made a few comments regarding whether we should keep the meta.yaml and heavyai.conf files.

.conda/meta.yaml
tupui: It's only used in some release scripts but if you are never running these we can remove this and the scripts.

.conda/meta.yaml
tupui: Mistake on my side, I looked at the wrong meta file (previous release) when I started this and then forgot to update this one.

tupui (commented):
Thanks for the review @guilhermeleobas. I am for removing the conda recipe. The config files are only used by the docker jobs in the CI which are not run. I can also remove this if wanted.

I can re-add clang as a testing dependency. FYI it was failing with latest clang.

.conda/meta.yaml
guilhermeleobas: Is the script you mentioned `update-conda-pkg.sh`? If yes, the `meta.yaml` file it reads is the one from the `rbc-feedstock` repo. Also, I think we can remove this script as well since the conda release process is automatically triggered.

guilhermeleobas:
> FYI it was failing with latest clang.

We need to use the same version of llvm tools used by Numba. See:
https://github.com/xnd-project/rbc/issues/334

tupui:
Ah great, I will update all of that.

.github/workflows/rbc_test.yml
```diff
@@ -150,19 +150,6 @@ jobs:
...
-          HEAVYDB_DOCKER_IMAGE: ${{ matrix.docker-image }}
-        if: matrix.os == 'ubuntu-latest' && matrix.heavydb-from == 'docker'
-        run: |
-          # force docker-compose installation
-          mamba create -n docker docker-compose -c conda-forge
-          # run docker services
-          mamba run -n docker docker-compose build
-          mamba run -n docker docker-compose up -d
-          sleep 20
-
```
guilhermeleobas, line 165: Oh, sorry. Maybe I wasn't clear in my last comment. We should keep the instructions to run the heavydb docker image.

.github/workflows/rbc_test.yml
```diff
@@ -215,19 +202,6 @@ jobs:
...
-        env:
-          HEAVYDB_DEV_LABEL: ${{ 'docker-dev' }}
-          EXPECTED_PYTHON_VERSION: ${{ matrix.python-version }}
-          EXPECTED_HEAVYDB_VERSION: ${{ matrix.heavydb-version }}
-          EXPECTED_NUMBA_VERSION: ${{ matrix.numba-version }}
-          RBC_TESTS_FULL: TRUE
-        run: |
-          mamba run -n rbc pytest -sv -r A rbc/tests/ -x -k heavydb
-          mamba run -n rbc pytest -sv -r A rbc/tests/ -x -k "not heavydb"
-
```
guilhermeleobas, line 230: restore.

.github/workflows/rbc_test.yml
```diff
@@ -254,13 +228,6 @@ jobs:
             mamba run -n heavydb-env cat data/mapd_log/omnisci_server.ERROR
           fi
 
-      - name: Show Heavydb docker logs on failure [docker]
-        shell: bash -l {0}
-        if: failure() && matrix.os == 'ubuntu-latest' && matrix.heavydb-from == 'docker'
-        run: |
-          mamba run -n docker docker-compose top  # to check if server is still alive
-          mamba run -n docker docker-compose exec -T ls
-
```
guilhermeleobas, line 263: restore

.github/workflows/rbc_test.yml
```diff
@@ -279,14 +246,6 @@ jobs:
...
           cat heavydb-output.txt
 
-      - name: Show Heavydb docker logs on failure [docker]
-        shell: bash -l {0}
-        timeout-minutes: 2
-        if: ${{ failure() && matrix.os == 'ubuntu-latest' && matrix.heavydb-from == 'docker' }}
-        run: |
-          mamba run -n docker docker-compose logs --no-color --tail=10000 -f -t \> heavydb-docker.log
-          cat heavydb-docker.log
-
```
guilhermeleobas, line 289: restore

.omnisci/omnisci.conf
```diff
@@ -1,5 +0,0 @@
-enable-runtime-udf = true
-enable-table-functions = true
-enable-dev-table-functions = true
-cpu-only = true
-log-severity = DEBUG4
```
guilhermeleobas, line 5: Yeah, sorry about that. We should keep this file too.

.heavyai/heavyai.conf
```diff
@@ -1,5 +0,0 @@
-enable-runtime-udf = true
-enable-table-functions = true
-enable-dev-table-functions = true
-cpu-only = true
-log-severity = DEBUG4
```
guilhermeleobas, line 5: and this one!

environment.yml
```diff
@@ -13,12 +13,9 @@ dependencies:
   - python>=3.7
   - setuptools
   - numba>=0.52
```
guilhermeleobas, line 15:
```suggestion
  - numba>=0.54
```

environment.yml
```diff
@@ -13,12 +13,9 @@ dependencies:
...
-  - llvmlite
+  - llvmlite>=0.29
```
guilhermeleobas, line 16:
```suggestion
  - llvmlite
```

guilhermeleobas (changes requested):

tupui:
@guilhermeleobas I reverted the docker changes.

guilhermeleobas (approved):

PR #499, opened 2022-06-14
Add support for string proxy methods

guilhermeleobas:
* Add `string_dict_proxy` attribute to the type if it is a `Column(List)<TextEncodingDict>`. This is done using the `buffer_extra_members` method available in the RBC `Type` class.
* Expose the following methods to add and retrieve strings
  * `::getString(string_id)`
  * `::getStringId(string)`
* Enables `nrt` to support unicode strings. From a quick scan of the [`unicode.py`](https://github.com/numba/numba/blob/main/numba/cpython/unicode.py) file, it doesn't seem to require the Python API.
* Require https://github.com/heavyai/heavydb-internal/pull/6695


brenocfg (changes requested):
Suggested a discussion on `StringDictionaryProxy_getString()` that I think we should settle before approving. (The rest LGTM though)

rbc/heavydb/column.py
```diff
@@ -95,21 +104,80 @@ def codegen(context, builder, signature, args):
...
 
         return builder.icmp_signed('==', res, nv)
 
     return sig, codegen
 
 
+@extending.intrinsic
+def heavydb_column_getString_(typingctx, col_var, row_idx):
+    # should this return a text encoding none? or even a string?
+    sig = nb_types.voidptr(col_var, row_idx)
```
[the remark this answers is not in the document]
brenocfg: Also, shouldn't `col_var` be renamed to `dict_ptr` and `row_idx` to `string_id` or something of that sort? Just to make it clear what these two parameters actually refer to?

rbc/heavydb/column.py
```diff
@@ -95,21 +106,124 @@ def codegen(context, builder, signature, args):
...
+
+        ptr = getBytes(builder, proxy, string_id)
+        sz = getBytesLength(context, builder, proxy, string_id)
+
+        text = heavydb_buffer_constructor(context, builder, signature,
+                                          [builder.add(sz, sz.type(1))])
+        cgutils.memcpy(builder, text.ptr, ptr, sz)
+        builder.store(text.ptr.type.pointee(0), builder.gep(text.ptr, [sz]))
+        text.sz = sz
+        cgutils.printf(builder, "%s - %d\n", text.ptr, text.sz)
```
brenocfg, line 151: Leftover print I'm guessing? 😅

guilhermeleobas: Yes 😆

rbc/heavydb/column_list.py
```diff
@@ -18,20 +18,62 @@ class HeavyDBColumnListType(Type):
...
+def heavydb_columnlist_getitem(typingctx, lst, idx):
+    from .column import HeavyDBColumnType
+
+    members = lst.dtype.members
+    is_text_encoding_dict = False
+    if len(members) == 4:
+        is_text_encoding_dict = True
+        T = Type.fromstring('TextEncodingDict')
+    else:
+        T = Type.fromnumba(lst.dtype.members[0].dtype.dtype)
```
guilhermeleobas, line 51: Perhaps there's a better way to detect if the columnlist is of type `TextEncodingDict`.

rbc/heavydb/column.py
```diff
@@ -30,6 +32,12 @@ def match(self, other):
         if type(self) is type(other):
             return self[0] == other[0]
 
+    @property
+    def buffer_extra_members(self):
+        if self.element_type.tostring() == 'TextEncodingDict':
+            return ('i8* string_dict_proxy_',)
```
guilhermeleobas, line 38: note to myself: check if the heavydb version is >= (6, 2)

pearu (approved):

PR #500, opened 2022-06-23
Update docs

guilhermeleobas:


tupui:
For the documentation of dunder methods, you can try this https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html?highlight=special-members#directive-option-automodule-special-members

guilhermeleobas:
Is there a way to show dunder methods only for a specific class?

tupui:
> Is there a way to show dunder methods only for a specific class?

Yes, instead you need to add the directive just for the class. See https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html?highlight=special-members#directive-option-automodule-special-members

pearu (approved):
LGTM!

PR #501, opened 2022-06-23
Implement array.size and array.ndim attributes

guilhermeleobas:


rbc/heavydb/array.py
```diff
@@ -132,20 +126,34 @@ def T(self):
...
         else:
             raise errors.NumbaNotImplementedError(repr(dtype))
         return HeavyDBArrayType((element_type,)).tonumba()
     return typer
+
+
+@extending.overload_attribute(ArrayPointer, 'ndim')
+def get_ndim(arr):
+    def impl(arr):
+        return 1
```
tupui, line 151: Ah yes because we don't have matrix yet right?

tupui (approved):
LGTM, thanks @guilhermeleobas

rbc/heavydb/array.py
guilhermeleobas: Yes! Arrays in heavydb are single-dimension

pearu (approved):
LGTM!

PR #502, opened 2022-07-01
Numba 0.56

guilhermeleobas:
* Increase Numba supported version to 0.55
* Remove Numba 0.54 specific code


pearu (approved):
LGMT! Thanks, @guilhermeleobas!

PR #505, opened 2022-07-15
Fix an error in return buffer values on UDF

guilhermeleobas:
Currently, RBC uses the Numba calling convention and bet that LLVM can optmize the code to the point the result IR conforms to the HeavyDB calling convention. This pull request is the first patch to fix this, so we don't end up with unwanted `unreachable()` instructions.

### Calling convention

Given the function `fn` declared below which returns a copy of the input array. Numba will generate an LLVM IR which stores the input array pointer into the return pointer.

```python
def fn(a: Array) -> Array:
    ret = a
    return ret
```

```llvm
define i32 @fn__iner_cpu_0({ i64*, i64, i8 }** noalias nocapture %retptr, { i8*, i32, i8* }** noalias nocapture %excinfo, { i64*, i64, i8 }* %arg.a) {
entry:
  br label %B0

B0:                                               ; preds = %entry
  store { i64*, i64, i8 }* %arg.a, { i64*, i64, i8 }** %retptr, align 8
  ret i32 0
}
```

By the Numba calling convention

1) Return type is passed as the first argument to the function (`%retptr`)
2) Second argument is a pointer to an exception info struct. Basically, it holds if the function raised an exception or not. Since we don't support exceptions, this struct is not used and deleted by DCE.
3) Third argument is a pointer to the input array.

RBC also generates a wrapper function that conforms with the HeavyDB calling convention:

* If the UDF returns a scalar, its signature is `udf(arg1, ..., argN) -> scalar`
* If its a struct, then the signature is `udf(retptr*, arg1, ..., argN) -> void`

The problem with the wrapper we generate is that we only copy the return pointer from the inner function, whereas it should be deep copied.

```llvm
define void @"fn__wrapper_cpu_0"({i64*, i64, i8}* %".1", {i64*, i64, i8}* %".2"){
entry:
  %.4 = alloca { i64*, i64, i8 }*, align 8
  store { i64*, i64, i8 }* null, { i64*, i64, i8 }** %.4, align 8
  %excinfo = alloca { i8*, i32, i8* }*, align 8
  store { i8*, i32, i8* }* null, { i8*, i32, i8* }** %excinfo, align 8
  %.8 = call i32 @fn__inner_cpu_0({ i64*, i64, i8 }** %.4, { i8*, i32, i8* }** %excinfo, i64 %.2)
  %.9 = load { i8*, i32, i8* }*, { i8*, i32, i8* }** %excinfo, align 8
  %.10 = icmp eq i32 %.8, 0
  %.11 = icmp eq i32 %.8, -2
  %.14 = or i1 %.10, %.11
  %.15 = xor i1 %.14, true
  %.16 = icmp sge i32 %.8, 1
  %.18 = load { i64*, i64, i8 }*, { i64*, i64, i8 }** %.4, align 8
  br i1 %.15, label %entry.if, label %entry.endif, !prof !0

entry.if:                                         ; preds = %entry
  ret void

entry.endif:                                      ; preds = %entry
  %.21 = load { i64*, i64, i8 }, { i64*, i64, i8 }* %.18, align 8
  ; copies the return pointer from inner into the retptr (%.1)
  store { i64*, i64, i8 } %.21, { i64*, i64, i8 }* %.1, align 8
  ret void
}
```

This PR introduces a new method that Buffer subclasses can overload. This method is responsible for generating LLVM code that copies each buffer element into the return pointer.

### DIFF
--- a/rbc/heavydb/array.py
+++ b/rbc/heavydb/array.py
@@ -7,9 +7,14 @@ from rbc import typesystem, errors
 from .buffer import (BufferPointer, Buffer,
                      HeavyDBBufferType,
                      heavydb_buffer_constructor)
-from numba.core import extending
+from numba.core import extending, cgutils
 from numba import types as nb_types
 from typing import Union
+from llvmlite import ir
+
+
+int32_t = ir.IntType(32)
+int64_t = ir.IntType(64)
 
 
 class HeavyDBArrayType(HeavyDBBufferType):
@@ -24,12 +29,39 @@ class HeavyDBArrayType(HeavyDBBufferType):
         bool is_null;
       }
     """
+
+    @property
+    def numba_pointer_type(self):
+        return ArrayPointer
+
     @property
     def buffer_extra_members(self):
         return ('bool is_null',)
 
 
-ArrayPointer = BufferPointer
+class ArrayPointer(BufferPointer):
+    def copy(self, context, builder, val, retptr):
+        from .buffer import memalloc
+        ptr_type = self.dtype.members[0]
+        element_size = int64_t(ptr_type.dtype.bitwidth // 8)
+
+        src = builder.extract_value(builder.load(val), 0)
+        element_count = builder.extract_value(builder.load(val), 1)
+        is_null = builder.extract_value(builder.load(val), 2)
+
+        zero, one, two = int32_t(0), int32_t(1), int32_t(2)
+        with builder.if_else(cgutils.is_true(builder, is_null)) as (then, otherwise):
+            with then:
+                nullptr = cgutils.get_null_value(src.type)
+                builder.store(nullptr, builder.gep(retptr, [zero, zero]))
+            with otherwise:
+                # we can't just copy the pointer here because return buffers need
+                # to have their own memory, as input buffers are freed upon returning
+                ptr = memalloc(context, builder, ptr_type, element_count, element_size)
+                cgutils.raw_memcpy(builder, ptr, src, element_count, element_size)
+                builder.store(ptr, builder.gep(retptr, [zero, zero]))
+        builder.store(element_count, builder.gep(retptr, [zero, one]))
+        builder.store(is_null, builder.gep(retptr, [zero, two]))
 
 
 class Array(Buffer):
--- a/rbc/heavydb/array.py
+++ b/rbc/heavydb/array.py
@@ -40,7 +40,7 @@ class HeavyDBArrayType(HeavyDBBufferType):
 
 
 class ArrayPointer(BufferPointer):
-    def copy(self, context, builder, val, retptr):
+    def deepcopy(self, context, builder, val, retptr):
         from .buffer import memalloc
         ptr_type = self.dtype.members[0]
         element_size = int64_t(ptr_type.dtype.bitwidth // 8)
--- a/rbc/heavydb/array.py
+++ b/rbc/heavydb/array.py
@@ -45,9 +45,10 @@ class ArrayPointer(BufferPointer):
         ptr_type = self.dtype.members[0]
         element_size = int64_t(ptr_type.dtype.bitwidth // 8)
 
-        src = builder.extract_value(builder.load(val), 0)
-        element_count = builder.extract_value(builder.load(val), 1)
-        is_null = builder.extract_value(builder.load(val), 2)
+        struct_load = builder.load(val, name='struct_load')
+        src = builder.extract_value(struct_load, 0, name='array_buff_ptr')
+        element_count = builder.extract_value(struct_load, 1, name='array_size')
+        is_null = builder.extract_value(struct_load, 2, name='array_is_null')
 
         zero, one, two = int32_t(0), int32_t(1), int32_t(2)
         with builder.if_else(cgutils.is_true(builder, is_null)) as (then, otherwise):
--- a/rbc/heavydb/text_encoding_none.py
+++ b/rbc/heavydb/text_encoding_none.py
@@ -9,16 +9,17 @@ from rbc import typesystem
 from rbc.targetinfo import TargetInfo
 from rbc.errors import RequireLiteralValue
 from .buffer import (
-    BufferPointer, Buffer, HeavyDBBufferType,
+    Buffer, HeavyDBBufferType,
     heavydb_buffer_constructor)
 from numba.core import extending, cgutils
 from numba.core import types as nb_types
+from .array import ArrayPointer
 from llvmlite import ir
 from typing import Union
 
 
-class TextEncodingNonePointer(BufferPointer):
-    pass
+int32_t = ir.IntType(32)
+int64_t = ir.IntType(64)
 
 
 class HeavyDBTextEncodingNoneType(HeavyDBBufferType):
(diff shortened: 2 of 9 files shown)


rbc/heavydb/array.py
```diff
@@ -24,12 +29,39 @@ class HeavyDBArrayType(HeavyDBBufferType):
...
-ArrayPointer = BufferPointer
+class ArrayPointer(BufferPointer):
+    def copy(self, context, builder, val, retptr):
+        from .buffer import memalloc
+        ptr_type = self.dtype.members[0]
+        element_size = int64_t(ptr_type.dtype.bitwidth // 8)
+
+        src = builder.extract_value(builder.load(val), 0)
+        element_count = builder.extract_value(builder.load(val), 1)
+        is_null = builder.extract_value(builder.load(val), 2)
```
brenocfg, line 50: Sort of unrelated to this PR, but should we start naming our IR instructions? I think it makes the IR a lot more readable during debugging.

For instance, these could be:

```python
src = builder.extract_value(builder.load(val, "array_struct"), 0, "array_buff_ptr")
element_count = builder.extract_value(builder.load(val, "array_struct"), 1, "array_size")
is_null = builder.extract_value(builder.load(val, "array_struct"), 2, "array_is_null")
```

It would also help separate what is rbc-generated IR from numba-generated IR. I think as long as the instruction supports it, we should name it to reflect its counterpart in the code, so that the IR is close to as readable as the code that generates it.

Not only is it a bonus for readability, but I think it helps a lot in creating an "index" for generated IR instructions. E.g., if you're trying to look for which code generated some IR, you can CTRL+F for its name and usually find the corresponding builder call that emitted it.

brenocfg, line 50: I know numba/LLVM will optimize into this result anyway, but I think the code could be a bit more readable if we just emit/declare the redundant load separately:

```suggestion
        struct_load = builder.load(val)
        src = builder.extract_value(struct_load, 0)
        element_count = builder.extract_value(struct_load, 1)
        is_null = builder.extract_value(struct_load, 2)
```

rbc/heavydb/array.py
```diff
@@ -24,12 +29,39 @@ class HeavyDBArrayType(HeavyDBBufferType):
...
+
+        zero, one, two = int32_t(0), int32_t(1), int32_t(2)
+        with builder.if_else(cgutils.is_true(builder, is_null)) as (then, otherwise):
+            with then:
+                nullptr = cgutils.get_null_value(src.type)
+                builder.store(nullptr, builder.gep(retptr, [zero, zero]))
+            with otherwise:
+                # we can't just copy the pointer here because return buffers need
+                # to have their own memory, as input buffers are freed upon returning
+                ptr = memalloc(context, builder, ptr_type, element_count, element_size)
```
brenocfg, line 60: I know we discussed this before but I can't quite remember the answer. Who takes care of this memory once the UDF returns? Is it deallocated by heavydb's `Array<T>` destructor?

rbc/heavydb/text_encoding_none.py
```diff
@@ -42,8 +47,22 @@ def match(self, other):
...
-    pass
+class TextEncodingNonePointer(ArrayPointer):
+    def copy(self, context, builder, val, retptr):
+        from .buffer import memalloc
+        ptr_type = self.dtype.members[0]
+        element_size = int64_t(ptr_type.dtype.bitwidth // 8)
+
+        src = builder.extract_value(builder.load(val), 0)
+        element_count = builder.extract_value(builder.load(val), 1)
+        is_null = builder.extract_value(builder.load(val), 2)
```
brenocfg, line 58: Same as above wrt the redundant load.

brenocfg (changes requested):
Suggested a few readability/QoL changes, otherwise LGTM!

rbc/heavydb/array.py
guilhermeleobas: It is deallocated by heavydb

guilhermeleobas:
I was wondering if the changes introduced in this PR isn't the responsibility of heavydb. Because I think the same errors will happen in C++ UDFs as well.

rbc/heavydb/text_encoding_none.py
```diff
@@ -47,6 +48,25 @@ def match(self, other):
...
 
+class TextEncodingNonePointer(ArrayPointer):
+    def copy(self, context, builder, val, retptr):
+        from .buffer import memalloc
+        ptr_type = self.dtype.members[0]
+        element_size = int64_t(ptr_type.dtype.bitwidth // 8)
+
+        struct_load = builder.load(val)
+        src = builder.extract_value(struct_load, 0, name='text_buff_ptr')
+        element_count = builder.extract_value(struct_load, 1, name='text_size')
```
pearu, line 59: If `element_count` is negative, we could use `strlen` to compute the actual `element_count`?

rbc/heavydb/text_encoding_none.py
```diff
@@ -47,6 +48,25 @@ def match(self, other):
             return 2
 
 
+class TextEncodingNonePointer(ArrayPointer):
+    def copy(self, context, builder, val, retptr):
```
pearu, line 52: Suggestion: rename `copy` to `deepcopy`.

pearu (changes requested):
Thanks, @guilhermeleobas! Overall, this looks good.

I have a couple of nits and a concern with `element_count` value being negative for non-fixed-length TextEncodingNone cases (assuming that the doc string is correct). Could you add a test for this case as see what happens and if there is need to use `strlen` to recompute the `element_count`?

rbc/heavydb/text_encoding_none.py
```diff
@@ -47,6 +48,25 @@ def match(self, other):
...
+        element_size = int64_t(ptr_type.dtype.bitwidth // 8)
+
+        struct_load = builder.load(val)
+        src = builder.extract_value(struct_load, 0, name='text_buff_ptr')
+        element_count = builder.extract_value(struct_load, 1, name='text_size')
+        is_null = builder.extract_value(struct_load, 2, name='text_is_null')
+
+        zero, one, two = int32_t(0), int32_t(1), int32_t(2)
+        ptr = memalloc(context, builder, ptr_type, element_count, element_size)
+        cgutils.raw_memcpy(builder, ptr, src, element_count, element_size)
```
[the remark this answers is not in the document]
guilhermeleobas: I don't think we can have a text encoding none with negative size. Size should reflect the number of characters on `ptr_`. If such case happen, it is a bug.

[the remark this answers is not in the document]
guilhermeleobas: I think the docstring is wrong. At least on HeavyDB, the size member always represents the length of the string.

rbc/heavydb/array.py
```diff
@@ -24,12 +29,40 @@ class HeavyDBArrayType(HeavyDBBufferType):
...
```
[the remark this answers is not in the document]
guilhermeleobas: See my comment above. I think the docstring is wrong. `element_count` should never be negative.

rbc/heavydb/text_encoding_none.py
guilhermeleobas: See my comment above. `element_count` should never be negative.

pearu (approved):

PR #506, opened 2022-07-22
Maintainance of rbc for heavydb 6.2

pearu:
As in the title:
- update thrift files
- fix tests
- minor nits


rbc/tests/heavydb/test_heavydb_thrift.py
```diff
@@ -33,15 +33,15 @@ def foo(x, y):
 def test_get_udtf_details(heavydb):
     heavydb.unregister()
 
-    @heavydb('int32(Column<int>, OutputColumn<int>)')
+    @heavydb('int32(Column<int>, OutputColumn<int>)', devices=['cpu'])
```
guilhermeleobas, line 36: Can this UDTF run in the GPU?

guilhermeleobas (approved):
CI is all green. I guess the timeout issue was because of some problem in GA. Nevertheless, we can leave the timeout in 35 minutes.

rbc/tests/heavydb/test_heavydb_thrift.py
pearu: Yes, it can. This changeset was introduced to make the test stable: when both CPU and GPU would be enabled, then `len(details)` would be 2 and the assert statement below would fail.

PR #509, opened 2022-09-29
Add RowFunctionManager to Udfs

guilhermeleobas:
Requires: https://github.com/heavyai/heavydb-internal/pull/6961

### DIFF
--- a/rbc/extension_functions.thrift
+++ b/rbc/extension_functions.thrift
@@ -93,7 +93,7 @@ struct TUserDefinedFunction {
   1: string name,
   2: list<TExtArgumentType> argTypes,
   3: TExtArgumentType retType,
-  4: bool usesManager
+  4: list<map<string, string>> annotations
 }
 
 struct TUserDefinedTableFunction {
--- a/rbc/extension_functions.thrift
+++ b/rbc/extension_functions.thrift
@@ -91,8 +91,9 @@ enum TOutputBufferSizeType {
 
 struct TUserDefinedFunction {
   1: string name,
-  2: list<TExtArgumentType> argTypes
-  3: TExtArgumentType retType
+  2: list<TExtArgumentType> argTypes,
+  3: TExtArgumentType retType,
+  4: bool usesManager
 }
 
 struct TUserDefinedTableFunction {
--- a/rbc/heavydb/row_function_manager.py
+++ b/rbc/heavydb/row_function_manager.py
@@ -1,0 +1,131 @@
+__all__ = ['HeavyDBRowFunctionManagerType']
+
+
+from numba.core import extending, types, cgutils
+from numba.core.cgutils import make_bytearray, global_constant
+from rbc import structure_type
+from rbc.errors import UnsupportedError, RequireLiteralValue
+from rbc.targetinfo import TargetInfo
+from rbc.typesystem import Type
+from llvmlite import ir
+
+
+class HeavyDBRowFunctionManagerType(Type):
+    """RowFunctionManager<> is a typesystem custom type that
+    represents a class type with the following public interface:
+
+      struct RowFunctionManager { }
+
+    """
+
+    @property
+    def __typesystem_type__(self):
+        ptr_t = Type.fromstring("int8 ptr")
+        return Type(ptr_t).params(NumbaPointerType=HeavyDBRowFunctionManagerNumbaType).pointer()
+
+
+class HeavyDBRowFunctionManagerNumbaType(structure_type.StructureNumbaPointerType):
+    pass
+
+
+error_msg = 'RowFunctionManager is only available in HeavyDB 6.2 or newer (got %s)'
+i8p = ir.IntType(8).as_pointer()
+i32 = ir.IntType(32)
+i64 = ir.IntType(64)
+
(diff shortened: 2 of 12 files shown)


rbc/extension_functions.thrift
```diff
@@ -91,8 +91,9 @@ enum TOutputBufferSizeType {
 
 struct TUserDefinedFunction {
   1: string name,
-  2: list<TExtArgumentType> argTypes
-  3: TExtArgumentType retType
+  2: list<TExtArgumentType> argTypes,
+  3: TExtArgumentType retType,
+  4: bool usesManager
```
pearu, line 96: `TUserDefinedTableFunction` uses
```
list<map<string, string>> annotations
```
than enables introducing new parameters without changing the thrift interface.
I suggest using the same approach here, for instance, via:
```suggestion
  4: list<map<string, string>> annotations
```
so that the signature `<return value>(RowFunctionManager, <arg0>, <arg1>, ...)` is expressed as `annotations=[<arg0 annotations>, <arg1 annotations>, ..., <return value annotations>, dict(uses_manager='1')]` when following the same logic as in `TableFunction::getInputAnnotations`, `TableFunction::getOutputAnnotations`, and `TableFunction::getFunctionAnnotations`.

rbc/heavydb/row_function_manager.py
```diff
@@ -0,0 +1,132 @@
...
+    represents a class type with the following public interface:
+
+      struct RowFunctionManager { }
+
+    """
+
+    @property
+    def __typesystem_type__(self):
+        ptr_t = Type.fromstring("int8 pointer")
+        return Type(ptr_t).params(NumbaPointerType=HeavyDBRowFunctionManagerNumbaType).pointer()
```
pearu, line 25: `ptr_t == int8` seems misleading to me. Suggest 
```suggestion
        return Type('int8').params(NumbaPointerType=HeavyDBRowFunctionManagerNumbaType).pointer()
```

rbc/tests/heavydb/test_text_encoding_dict.py
```diff
@@ -488,3 +494,26 @@ def test_getString_lst(heavydb, col):
         assert list(result) == [(9,), (9,), (9,), (12,), (15,)]
     else:
         assert list(result) == [(18,), (9,), (9,), (9,)]
+
+
+@pytest.mark.parametrize('col', ['t1'])
+def test_udf_copy_dict_encoded_string(heavydb, col):
+    if heavydb.version[:2] < (6, 2):
+        pytest.skip('Requires HeavyDB main branch')
```
pearu, line 502:
```suggestion
        pytest.skip('Requires HeavyDB version 6.2 or newer')
```

rbc/tests/heavydb/test_text_encoding_dict.py
```diff
@@ -488,3 +494,26 @@ def test_getString_lst(heavydb, col):
...
+    table = f"{heavydb.base_name}text"
+    query = f"SELECT fn_copy({col}) from {table}"
+    _, result = heavydb.sql_execute(query)
+
+    assert list(result) == [('fun',), ('bar',), ('foo',), ('barr',), ('foooo',)]
+
+
+def test_row_function_manager(heavydb):
+    if heavydb.version[:2] < (6, 3):
+        pytest.skip('Requires HeavyDB main branch')
```
pearu, line 513:
```suggestion
    if heavydb.version[:2] < (6, 2):
        pytest.skip('Requires HeavyDB version 6.2 or newer')
```

rbc/tests/heavydb/test_text_encoding_dict.py
```diff
@@ -488,3 +494,26 @@ def test_getString_lst(heavydb, col):
...
+
+    @heavydb('int32(int32, RowFunctionManager)')
+    def invalid_fn(a, mgr):
+        return a
```
pearu, line 517: Did you mean adding this within a try-except block?

pearu (changes requested):
I have a couple of nits and a suggestion to introduce `annotations` member to `TUserDefinedFunction` in order to (i) follow the convention used in `TUserDefinedTableFunction` and (ii) reduce the need to change the thrift interface in the future in case we will want to pass other parameters to the server.

rbc/extension_functions.thrift
guilhermeleobas: This requires updating the thrift in heavydb first, as I implemented with `bool usesManager` there.

rbc/tests/heavydb/test_text_encoding_dict.py
guilhermeleobas: Yes! Fixed

rbc/extension_functions.thrift
pearu: Could we change it there? We likely need `annotations` for UDFs anyway, for instance, for naming arguments etc.

guilhermeleobas: Yes, I'm doing the necessary changes there as well

pearu (approved):
LGMT! Thanks, @guilhermeleobas !

PR #512, opened 2022-10-14
Implement DayTimeInterval and YearMonthTimeInterval types

guilhermeleobas:


rbc/typesystem.py
```diff
@@ -726,6 +726,14 @@ def __str__(self):
             return self.tostring()
         return tuple.__str__(self)
 
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, self.__class__):
+            return False
+        return super().__eq__(other)
+
+    def __hash__(self) -> int:
+        return super().__hash__()
```
pearu, line 735: Why do we need to define `__hash__` here while the base class (tuple) already provides it?

pearu (approved):

rbc/typesystem.py
guilhermeleobas: Without it, the following fails:

```python
(Pdb++) hash(Type('float64'))
*** TypeError: unhashable type: 'Type'
```

PR #513, opened 2022-10-14
Support `Column<Array<T>>`

guilhermeleobas:
Requires https://github.com/heavyai/heavydb-internal/pull/7049

Add bindings for the following functions:
- `Column<Array<T>>::getitem`
- `Column<Array<T>>::setItem`
- `Column<Array<T>>::concatItem`
- `Column<Array<T>>::isNull`
- `Column<Array<T>>::setNull`
- `Column<Array<T>>::setNull`
- `TableFunctionManager::set_output_array_values_total_number`

### DIFF
--- a/rbc/heavydb/column_array.py
+++ b/rbc/heavydb/column_array.py
@@ -8,11 +8,16 @@ UDTFs.
 __all__ = ['HeavyDBOutputColumnArrayType', 'HeavyDBColumnArrayType']
 
 import operator
+from typing import TypeVar
 from rbc import typesystem
-from .column import HeavyDBColumnType
+from .column import HeavyDBColumnType, Column
+from .array import Array
 from numba.core import extending, cgutils
 from numba.core import types as nb_types
 from llvmlite import ir
+
+
+T = TypeVar('T')
 
 
 i1 = ir.IntType(1)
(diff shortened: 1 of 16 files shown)


rbc/heavydb/column_array.py
```diff
@@ -0,0 +1,352 @@
...
+        """
+        Set the array at ``index`` to null.
+
+        .. note::
+            Only available on ``CPU``
+        """
+
+    def __len__(self) -> int:
+        """
+        """
```
pearu, line 92:
```suggestion
        """
        Return the length of the array.
        
        ..  note::
             Only available on ``CPU``
        """
```

rbc/heavydb/column_array.py
```diff
@@ -0,0 +1,352 @@
...
+    the input and output arguments of UDTFs.
+    """
+
+
+class ColumnArrayType(nb_types.Type):
+    """Numba type class for HeavyDB buffer structures.
+    """
+
+    def __init__(self, dtype):
+        array_T = self.__typesystem_type__._params.get(_COLUMN_PARAM_NAME)
```
pearu, line 134: IIUC, `self` is `numba.core.types.Type` derived instance. Where is defined its `__typesystem_type__` attribute? In numba?

pearu (approved):
LGTM! Nice work, @guilhermeleobas!

I have one question though, see below.

rbc/heavydb/column_array.py
guilhermeleobas: Here:
https://github.com/xnd-project/rbc/blob/c576db2714c950b31bf799590d91bd42df768b57/rbc/structure_type.py#L93

PR #519, opened 2022-11-17
Update thrift files and bump server version to 6.4

guilhermeleobas:
Follow up of #518, which updates the server version to 6.4 and add new type definitions to the internal type map.


pearu (approved):

PR #521, opened 2022-12-14
Emit warning message when version(Numba LLVM) != version(clang)

guilhermeleobas:
Add a warning message when there's a version mismatch between the LLVM version shipped with Numba and the clang version selected


pearu (approved):

A warning: here we use the private numba symbol `_min_llvm_version`. Would it make sense to move the import statement under the `compiler` method?

PR #525, opened 2023-01-24
add to_text_encoding_none test

guilhermeleobas:
As title.

### DIFF
--- a/rbc/tests/heavydb/test_text_encoding_dict.py
+++ b/rbc/tests/heavydb/test_text_encoding_dict.py
@@ -120,11 +120,7 @@ def define(heavydb):
             db_id = mgr.getDictDbId('to_text_encoding_none_1', 0)
             dict_id = mgr.getDictId('to_text_encoding_none_1', 0)
             str = mgr.getString(db_id, dict_id, t)
-            n = len(str)
-            r = TextEncodingNone(n)
-            for i in range(n):
-                r[i] = str[i]
-            return r
+            return str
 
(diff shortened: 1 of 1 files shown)


rbc/tests/heavydb/test_text_encoding_dict.py
```diff
@@ -114,6 +115,28 @@ def fn_copy(mgr, t):
...
+        @heavydb('TextEncodingNone(RowFunctionManager, TextEncodingDict)', devices=['cpu'])
+        def to_text_encoding_none_1(mgr, t):
+            db_id = mgr.getDictDbId('to_text_encoding_none_1', 0)
+            dict_id = mgr.getDictId('to_text_encoding_none_1', 0)
+            str = mgr.getString(db_id, dict_id, t)
+            n = len(str)
+            r = TextEncodingNone(n)
+            for i in range(n):
+                r[i] = str[i]
+            return r
```
pearu, line 127: Why there is two UDFs with identical implementations?
Did you mean:
```suggestion
            return str
```
?

pearu (changes requested):
Thanks for the PR, @guilhermeleobas!

rbc/tests/heavydb/test_text_encoding_dict.py
guilhermeleobas: Oops, my bad. Fixed

pearu (commented):
LGTM!

PR #527, opened 2023-01-25
Add instructions to build and test RBC

guilhermeleobas:
Probably depends on #526

### DIFF
--- a/rbc/heavydb/column_list.py
+++ b/rbc/heavydb/column_list.py
@@ -91,6 +91,7 @@ def heavydb_columnlist_getitem(typingctx, lst, idx):
 class ColumnList(Buffer):
     """
     RBC ``ColumnList<T>`` type that corresponds to HeavyDB COLUMN LIST
+    of arrays
 
     In HeavyDB, a ColumnList of type ``T`` is represented as follows:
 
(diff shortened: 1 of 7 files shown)


rbc/heavydb/column_list.py
```diff
@@ -91,6 +91,7 @@ def codegen(context, builder, signature, args):
 class ColumnList(Buffer):
     """
     RBC ``ColumnList<T>`` type that corresponds to HeavyDB COLUMN LIST
+    of arrays
```
pearu, line 94: hmm, why arrays? It could be any of any support types.

rbc/heavydb/column_list_array.py
```diff
@@ -30,6 +30,17 @@
 
 class ColumnListArray(ColumnList):
     """
+    RBC ``ColumnList<Array<T>>`` type that corresponds to HeavyDB COLUMN LIST
```
pearu, line 33:
```suggestion
    RBC ``ColumnList<Array<T>>`` type that corresponds to HeavyDB COLUMN LIST of arrays.
```

rbc/heavydb/column_list.py
pearu: perhaps this was meant for ColumnListArray doc-string..

pearu (changes requested):
I have a couple of nits, otherwise looks good!
Thanks, @guilhermeleobas!

rbc/heavydb/column_list.py
guilhermeleobas: Yes, ColumnListArray

pearu (approved):

PR #529, opened 2023-02-06
Introduce Opaque pointers

guilhermeleobas:
This is a refactor PR:

* Introduces an Opaque pointer type for types that are just a pointer to an actual implementation in HeavyDB
    * TableFunctionManager
    * RowFunctionManager
    * StringDictionaryProxy
* Simplify overload implementations using `external`


rbc/external.py
```diff
@@ -38,7 +38,7 @@ def external(cls, *args):
                     f"external function name not specified for signature {signature}"
                 )
 
-            for device in [a for a in t.annotation() or [] if a in ["CPU", "GPU"]] or [
+            for device in [a.upper() for a in t.annotation() or [] if a in ["CPU", "GPU"]] or [
```
pearu, line 41:
```suggestion
            for device in [a.upper() for a in t.annotation() or [] if a.upper() in ["CPU", "GPU"]] or [
```
?

rbc/heavydb/row_function_manager.py
```diff
@@ -57,121 +60,88 @@ def getDictId(self, func_name: str, arg_idx: int) -> int:
...
+    if target_info.software[1][:3] < (6, 2, 0):
+        raise UnsupportedError(error_msg % (".".join(map(str, target_info.software[1]))))
 
+    i8p = nb_types.CPointer(nb_types.int8)
+    i32 = nb_types.int32
+    proxy = string_dict_proxy.StringDictionaryProxyNumbaType()
+    sig = typing.signature(proxy, i8p, i32, i32)
+
+    symbol = 'RowFunctionManager_getStringDictionaryProxy'
+    get_string_dict_proxy_ = nb_types.ExternalFunction(symbol, sig)
```
pearu, line 107: Below we have
```python
def heavydb_column_getString(proxy, string_id):
    getBytes = external('int8* StringDictionaryProxy_getStringBytes(int8*, int32)')
    getBytesLength = external('int64 StringDictionaryProxy_getStringLength(int8*, int32)')
    ...
```
for instance. The use of `external` looks cleaner. Can we use it here?

rbc/heavydb/table_function_manager.py
```diff
@@ -63,109 +67,58 @@ def set_output_array_values_total_number(self, index: int,
...
-        msg_global_var = global_constant(
-            builder.module,
-            "table_function_manager_error_message",
-            msg_const)
-        msg_ptr = builder.bitcast(msg_global_var, i8p)
-
-        fnty = ir.FunctionType(i32, [i8p, i8p])
-        fn = cgutils.get_or_insert_function(
-            builder.module, fnty, "TableFunctionManager_error_message")
+    defn = 'int32_t TableFunctionManager_error_message(int8_t*, int8_t*)|CPU'
```
pearu, line 81: Should we use `|CPU` suffix elsewhere as well?

pearu (changes requested):
I have a couple of questions.

rbc/external.py
guilhermeleobas: What about introducing a second argument to external to specify the device? Similar to how it is done in the `@udtf` decorator:

```python
fn = external(sig, device='cpu') # or 'gpu' / or both?
```

rbc/heavydb/row_function_manager.py
guilhermeleobas: In this particular case, no. A string dictionary proxy cannot be an argument of a UDF/UDTF and thus, not initialized. Attempting to do `external('int8* func(StringDictionaryProxy*')` will actually create a new Numba type inheriting from `numba.type.Type`. This was one of the corner cases I had to deal when implementing this function.

rbc/heavydb/table_function_manager.py
guilhermeleobas: Yes! fixed

rbc/external.py
pearu: Device hint is currently specified as an annotation to a signature. So, introducing `device` to `external` requires that `external` will check if it will conflict with the `device` hint in the signature.

I agree that specifying the device in the signature is kind of a hack and we should try to move away from using this construct. So, yes, I think your suggestion makes sense.

In udtf decorator, we use `devices` API, so:
```
fn = external(sig, devices=['cpu'])
```

pearu (approved):

pearu:
This PR introduces the rbc tests failures when using cuda-enabled heavydb as reported in https://github.com/xnd-project/rbc/pull/537#issuecomment-1535011455

PR #530, opened 2023-02-10
Fix a bug when clang compiler is not available.

pearu:
As in the title plus improve warning message from UnsupporterError and prefer looking for a clang compiler with a numba specific llvm version.


guilhermeleobas (approved):
LGTM. Thanks @pearu

PR #531, opened 2023-02-17
Support Numba Runtime in RBC

guilhermeleobas:
This PR rewrites the Numba Runtime using LLVM.

### DIFF
--- a/rbc/heavydb/text_encoding_none.py
+++ b/rbc/heavydb/text_encoding_none.py
@@ -198,11 +198,21 @@ def heavydb_text_encoding_none_constructor_memcpy(context, builder, sig, args):
     return fa._getpointer()
 
 
+@extending.lower_builtin(TextEncodingNone, nb_types.UnicodeType)
+def text_encoding_none_unicode_ctor(context, builder, sig, args):
+
+    text = context.make_helper(builder, TextEncodingNonePointer)
+    # fn = get_copy_bytes_fn(builder.module, arr, src_data, sz)
+    text.sz = text.sz.type(0)
+    text.is_null = text.is_null.type(1)
+    return text
+
+
 @extending.type_callable(TextEncodingNone)
 def type_heavydb_text_encoding_none(context):
     def typer(sz):
         if isinstance(sz, nb_types.UnicodeType):
-            raise RequireLiteralValue('Requires StringLiteral')
+            return typesystem.Type.fromobject('TextEncodingNone').tonumba()
         if isinstance(sz, (nb_types.Integer, nb_types.StringLiteral)):
             return typesystem.Type.fromobject('TextEncodingNone').tonumba()
     return typer
@@ -267,4 +277,11 @@ def ol_attr_sz(text):
 def ol_attr_is_null(text):
     def impl(text):
         return ol_attr_is_null_(text)
+    return impl
+
+
+@extending.overload_method(TextEncodingNonePointer, 'to_string')
+def ol_to_string(text):
+    def impl(text):
+        return 'abcd'
     return impl
--- a/rbc/irtools.py
+++ b/rbc/irtools.py
@@ -10,10 +10,12 @@ import llvmlite.binding as llvm
 from .targetinfo import TargetInfo
 from .errors import UnsupportedError
 from . import libfuncs
+from rbc.nrt import create_nrt_functions
 from rbc.externals import stdio
 from numba.core import codegen, cpu, compiler_lock, \
     registry, typing, compiler, sigutils, cgutils, \
     extending, imputils
+from numba.core.runtime.context import NRTContext
 from numba.core import errors as nb_errors
 
 
@@ -157,6 +159,46 @@ class JITRemoteTypingContext(typing.Context):
         super().load_additional_registries()
 
 
+class JITRemoteNRTContext():
+    def __init__(self, context, enabled):
+        ...
+        # super().__init__(context, enabled)
+        # In Numba 0.57, we can redefine methods with the following table:
+        # _NRT_MEMINFO_HEAVY_API = _NRT_Meminfo_Functions(
+        #     "NRT_MemInfo_alloc_safe",
+        #     "NRT_MemInfo_alloc_dtor_safe",
+        #     "NRT_MemInfo_alloc_safe_aligned")
+        # self._meminfo_api = _NRT_MEMINFO_HEAVY_API
+
+    def incref(self, builder, typ, value):
+        ...
+
+    def decref(self, builder, typ, value):
+        ...
+
+    def meminfo_alloc_unchecked(self, builder, size):
+        """
+        Allocate a new MemInfo with a data payload of `size` bytes.
+
+        A pointer to the MemInfo is returned.
+
+        Returns NULL to indicate error/failure to allocate.
+        """
+        import pdb; pdb.set_trace()
+        self._require_nrt()
+
+        mod = builder.module
+        fnty = ir.FunctionType(cgutils.voidptr_t, [cgutils.intp_t])
+        fn = cgutils.get_or_insert_function(mod, fnty,
+                                            self._meminfo_api.alloc)
+        fn.return_value.add_attribute("noalias")
+        return builder.call(fn, [size])
+
+
+    def define_nrt_allocate_meminfo_and_data(self, builder, module):
+        pass
+
+
 class JITRemoteTargetContext(cpu.CPUContext):
 
     @compiler_lock.global_compiler_lock
@@ -165,6 +207,10 @@ class JITRemoteTargetContext(cpu.CPUContext):
         self.address_size = target_info.bits
         self.is32bit = (self.address_size == 32)
         self._internal_codegen = JITRemoteCodegen("numba.exec")
+
+    # @property
+    # def nrt(self):
+    #     return JITRemoteNRTContext(self, self.enable_nrt)
 
     def load_additional_registries(self):
         self.install_registry(imputils.builtin_registry)
@@ -381,11 +427,14 @@ def compile_to_LLVM(functions_and_signatures,
     typing_context = JITRemoteTypingContext()
     target_context = JITRemoteTargetContext(typing_context)
 
+    create_nrt_functions()
+
     # Bring over Array overloads (a hack):
     target_context._defns = target_desc.target_context._defns
 
     with replace_numba_internals_hack():
         codegen = target_context.codegen()
+        print(target_context)
         main_library = codegen.create_library('rbc.irtools.compile_to_IR')
         main_module = main_library._final_module
 
--- a/rbc/irtools.py
+++ b/rbc/irtools.py
@@ -38,7 +38,13 @@ def get_called_functions(library, funcname=None):
     for block in func.blocks:
         for instruction in block.instructions:
             if instruction.opcode == 'call':
-                name = list(instruction.operands)[-1].name
+                instr = list(instruction.operands)[-1]
+                name = instr.name
+                # ignore in a call to a function pointer
+                if 'bitcast' in str(instr):
+                    continue
+                if name == '':
+                    continue
                 f = module.get_function(name)
                 if name.startswith('llvm.'):
                     result['intrinsics'].add(name)
@@ -427,14 +433,13 @@ def compile_to_LLVM(functions_and_signatures,
     typing_context = JITRemoteTypingContext()
     target_context = JITRemoteTargetContext(typing_context)
 
-    create_nrt_functions()
+    nrt_module = create_nrt_functions(debug=debug)
 
     # Bring over Array overloads (a hack):
     target_context._defns = target_desc.target_context._defns
 
     with replace_numba_internals_hack():
         codegen = target_context.codegen()
-        print(target_context)
         main_library = codegen.create_library('rbc.irtools.compile_to_IR')
         main_module = main_library._final_module
 
@@ -443,6 +448,8 @@ def compile_to_LLVM(functions_and_signatures,
                 user_defined_llvm_ir = llvm.parse_assembly(user_defined_llvm_ir)
             assert isinstance(user_defined_llvm_ir, llvm.ModuleRef)
             main_module.link_in(user_defined_llvm_ir, preserve=True)
+
+        main_library.add_ir_module(nrt_module)
 
         succesful_fids = []
         function_names = []
--- a/rbc/nrt.py
+++ b/rbc/nrt.py
@@ -1,135 +1,454 @@
 from numba.core import cgutils
 from llvmlite import ir
+from contextlib import contextmanager
 
 void = ir.VoidType()
 i8 = ir.IntType(8)
 i8p = i8.as_pointer()
 i8pp = i8p.as_pointer()
+i32 = ir.IntType(32)
 i64 = ir.IntType(64)
 
 # { %"struct.std::atomic", void (i8*, i64, i8*)*, i8*, i8*, i64, i8* }
 # %"struct.std::atomic" = type { %"struct.std::__atomic_base" }
 # %"struct.std::__atomic_base" = type { i64 }
 
-atomic_i64 = ir.LiteralStructType([i64])
-meminfo_t = ir.LiteralStructType([atomic_i64, i8p, i8p, i8p, i64, i8p])
-meminfo_ptr_t = meminfo_t.as_pointer()
+REFCT_IDX = 0
+DTOR_IDX = 1
+DTOR_INFO_IDX = 2
+DATA_IDX = 3
+SIZE_IDX = 4
+EXTERNAL_ALLOCATOR_IDX = 5
+
+# struct MemInfo {
+#     std::atomic_size_t     refct;
+#     NRT_dtor_function dtor;
+#     void              *dtor_info;
+#     void              *data;
+#     size_t            size;    /* only used for NRT allocated memory */
+#     void* external_allocator;
+# };
+
+# atomic_i64_t is the llvm representation of std::atomic<int64>, according to godbolt.com
+atomic_i64_t = ir.global_context.get_identified_type('atomic_i64')
+atomic_i64_t.set_body(i64)
+
+# MemInfo struct
+MemInfo_t = ir.global_context.get_identified_type('Struct.MemInfo')
+# First argument ought to be "atomic_i64_t" instead of i64
+MemInfo_t.set_body(i64, i8p, i8p, i8p, i64, i8p)
+MemInfo_ptr_t = MemInfo_t.as_pointer()
+
+sizeof_MemInfo_t = i64(48)
 
 
+# Function Names
 NRT_MemInfo_init = 'NRT_MemInfo_init'
+
 NRT_MemInfo_alloc = 'NRT_MemInfo_alloc'
+NRT_MemInfo_alloc_safe = 'NRT_MemInfo_alloc_safe'
+
 NRT_MemInfo_alloc_dtor = 'NRT_MemInfo_alloc_dtor'
+NRT_MemInfo_alloc_dtor_safe = 'NRT_MemInfo_alloc_dtor_safe'
+
+NRT_Allocate = 'NRT_Allocate'
 NRT_Allocate_External = 'NRT_Allocate_External'
 NRT_MemInfo_call_dtor = 'NRT_MemInfo_call_dtor'
-nrt_alloc_meminfo_and_data = 'nrt_alloc_meminfo_and_data'
+
+NRT_incref = 'NRT_incref'
+NRT_decref = 'NRT_decref'
+
+NRT_MemInfo_data_fast = 'NRT_MemInfo_data_fast'
+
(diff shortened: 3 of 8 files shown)


rbc/nrt.py
```diff
@@ -0,0 +1,552 @@
...
+atomic_i64_t = ir.global_context.get_identified_type('atomic_i64')
+atomic_i64_t.set_body(i64)
+
+# MemInfo struct
+MemInfo_t = ir.global_context.get_identified_type('Struct.MemInfo')
+# First argument ought to be "atomic_i64_t" instead of i64
+MemInfo_t.set_body(i64, i8p, i8p, i8p, i64, i8p)
+MemInfo_ptr_t = MemInfo_t.as_pointer()
+
+sizeof_MemInfo_t = i64(48)
```
guilhermeleobas, line 44: This should be computed using target context

rbc/nrt.py
```diff
@@ -0,0 +1,552 @@
...
+# };
+
+# atomic_i64_t is the llvm representation of std::atomic<int64>, according to godbolt.com
+atomic_i64_t = ir.global_context.get_identified_type('atomic_i64')
+atomic_i64_t.set_body(i64)
+
+# MemInfo struct
+MemInfo_t = ir.global_context.get_identified_type('Struct.MemInfo')
+# First argument ought to be "atomic_i64_t" instead of i64
+MemInfo_t.set_body(i64, i8p, i8p, i8p, i64, i8p)
```
guilhermeleobas, line 41: Replace the first `i64` by `atomic_i64_t`

guilhermeleobas, line 41: `size_t` ought to be computed using target extension and target info

rbc/nrt.py
```diff
@@ -0,0 +1,552 @@
...
+            not_null = cgutils.is_not_null(builder, dtor)
+            with cgutils.if_likely(builder, not_null):
+                info_ = NULL
+                builder.call(dtor, [ptr, size, info_])
+        builder.ret_void()
+
+    def define_nrt_allocate_meminfo_and_data(self, builder, args):
+        with self.nrt_debug_ctx():
+            [size, mi_out, allocator] = args
+            # TODO: replace this by `get_abi_sizeof(...)`
```
guilhermeleobas, line 414: Address this TODO

rbc/nrt.py
```diff
@@ -0,0 +1,552 @@
...
+                with then:
+                    bb_then = builder.basic_block
+                    mi_ptr = builder.bitcast(ptr, MemInfo_ptr_t, name='meminfo_ptr')
+                    data = self._get_from_meminfo(mi_ptr, 'data')
+                    self.NRT_Debug("data: %p\n", data)
+                with otherwise:
+                    # there is a specific case where ptr is NULL and doesn't
+                    # crash Numba
+                    bb_else = builder.basic_block
+                    data_null = NULL
```
guilhermeleobas, line 329: Document this odd corner case

rbc/tests/heavydb/test_nrt_list.py
```diff
@@ -0,0 +1,112 @@
...
+    # [X] List.count
+    # [X] List.sort
+    # [X] List.reverse
+    # [X] List.copy
+
+    @heavydb("i32(i32)", devices=['cpu'])
+    def list_append(t):
+        lst = list()
+        lst.append('abc')
+        return len(lst)
```
guilhermeleobas, line 37: Not related to this test, but introduce an `array.to_list()` and `array.to_set()` methods

rbc/tests/heavydb/test_nrt_string.py
```diff
@@ -0,0 +1,277 @@
...
+        elif method == 'title':
+            r = s.title()
+        else:
+            r = ''
+        return TextEncodingNone(r)
+
+    @heavydb('bool(TextEncodingNone, TextEncodingNone, TextEncodingNone)')
+    def test_endswith(method, t, value):
+        s = t.to_string()
+        w = value.to_string()
```
guilhermeleobas, line 56: Introduce a test to ensure `to_string()` method is working correctly.

rbc/unicodetype_db.ll
```diff
@@ -0,0 +1,2297 @@
+
```
guilhermeleobas, line 1: Document how this file was generated and how to do it in the future if needed.

rbc/nrt.py
```diff
@@ -0,0 +1,552 @@
...
+        self.define()
+        self.set_hashsecrets()
+
+        if self.verbose:
+            print(self.module)
+
+    def set_hashsecrets(self):
+        for v in hashsecret.values():
+            gv = ir.GlobalVariable(self.module, i64, v.symbol)
+            gv.initializer = i64(v.value.value)
```
guilhermeleobas, line 167: hashsecrets should go to `unicodetype_db.ll`

rbc/nrt.py
```diff
@@ -0,0 +1,552 @@
...
+realloc = 'realloc'
+
+nrt_debug_incr = 'nrt_debug_incr'
+nrt_debug_decr = 'nrt_debug_decr'
+nrt_debug_spaces = 'nrt_debug_spaces'
+nrt_global_var = '__nrt_global_var'
+
+function_types = {
+    NRT_MemInfo_init: (ir.FunctionType(void, [MemInfo_ptr_t, i8p, i64, i8p, i8p, i8p]),
+                       ('mi', 'data', 'size', 'dtor', 'dtor_info', 'external_allocator')),
```
guilhermeleobas, line 90: Some of the `i64` values here are actually `size_t`. Replace it by the correct definition.

guilhermeleobas (commented):
Notes to myself

rbc/nrt.py
```diff
@@ -0,0 +1,552 @@
...
+        # allocator is always null as we don't use this argument in RBC/HeavyDB
+        with self.nrt_debug_ctx():
+            [size, allocator_] = args
+
+            ptr = builder.call(self.malloc, [size])
+            # allocate_varlen_buffer calls malloc using the formula:
+            #   malloc((elem_count + 1) * elem_size)
+            # elem_count = builder.sub(size, i64(1), name='element_count')
+            # elem_size = i64(1)
+            # ptr = builder.call(self.allocate_varlen_buffer, [elem_count, elem_size])
```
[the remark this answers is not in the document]
guilhermeleobas: Changed in [[1305c4a](https://github.com/xnd-project/rbc/pull/531/commits/1305c4a57a5d0c1106572e8c78d220ef3b809f35)](https://github.com/xnd-project/rbc/pull/531/commits/1305c4a57a5d0c1106572e8c78d220ef3b809f35)

rbc/heavydb/text_encoding_none.py
```diff
@@ -198,12 +182,58 @@ def heavydb_text_encoding_none_constructor_memcpy(context, builder, sig, args):
...
+            cgutils.raw_memcpy(builder, dst=p2, src=uni_str.data,
+                               count=uni_str.length, itemsize=i64(1))
+            # string is null terminated
+            builder.store(i8(0), builder.gep(p2, [uni_str.length]))
+            n2 = i8(0)
+
+    # {i8*, i64, i8}
+    #  ^^^
+    ptr = builder.phi(i8p)
+    ptr.add_incoming(p1, bb_then)
```
guilhermeleobas, line 218: No need to create `p1`
```suggestion
    ptr.add_incoming(NULL, bb_then)
```

rbc/heavydb/array.py
```diff
@@ -168,6 +174,19 @@ def heavydb_array_constructor(context, builder, sig, args):
...
+@extending.lower_builtin(Array, nb_types.List)
+def heavydb_array_ctor_list(context, builder, sig, args):
+    dtype = sig.args[0].dtype
+
+    def ctor(lst):
+        sz = len(lst)
+        arr = Array(sz, dtype)
+        for i in range(sz):
+            arr[i] = lst[i]
+        return arr
```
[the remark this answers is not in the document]
brenocfg: Depending on how complicated the IR becomes, there's a chance that LLVM will optimize it on its own with `-memcpyopt`: https://llvm.org/docs/Passes.html#memcpyopt-memcpy-optimization

rbc/heavydb/text_encoding_none.py
```diff
@@ -141,51 +152,24 @@ def heavydb_text_encoding_none_constructor(context, builder, sig, args):
...
+    module = builder.module
 
-    # create function and prevent llvm from optimizing it
-    fnty = ir.FunctionType(ir.VoidType(), [arr.type, src_data.type, sz.type])
-    func = ir.Function(module, fnty, fn_name)
-    func.attributes.add('noinline')
+    name = 'copy_bytes_fn'
+    fnty = ir.FunctionType(void, [i8p, i8p, i64])
+    fn = cgutils.get_or_insert_function(module, fnty, name)
+    fn.linkage = 'linkonce'
```
brenocfg, line 162: Why `linkonce` here? Can we have multiple definitions of this function in a module?

rbc/heavydb/text_encoding_none.py
```diff
@@ -198,12 +182,58 @@ def heavydb_text_encoding_none_constructor_memcpy(context, builder, sig, args):
     return fa._getpointer()
 
 
+@extending.lower_builtin(TextEncodingNone, nb_types.UnicodeType)
+def text_encoding_none_unicode_ctor(context, builder, sig, args):
+    # Note to the future:
+    # You shall be tempted to use "context.make_helper",
```
brenocfg, line 189: You said this but used `context.make_helper` in the very first following line of code 😅. I get why `extract_value` is a bad idea, but why the other two? It's not very clear what they do on first glance.

rbc/heavydb/text_encoding_none.py
```diff
@@ -198,12 +182,58 @@ def heavydb_text_encoding_none_constructor_memcpy(context, builder, sig, args):
...
+
+    eq = builder.icmp_signed('==', uni_str.length, uni_str.length.type(0))
+    with builder.if_else(eq, likely=False) as (then, otherwise):
+        with then:
+            bb_then = builder.basic_block
+            p1 = cgutils.get_null_value(i8p)
+            n1 = i8(1)
+        with otherwise:
+            bb_else = builder.basic_block
+            p2 = allocate_varlen_buffer(builder, uni_str.length, i64(1))
```
brenocfg, line 209: Is `uni_str.length` here the length of the string **including** the `\0` that terminates the string? Because in C `strlen` will usually be 1 byte short since it doesn't consider that:

```
sizeof("blah") = 5
strlen("blah") = 4
```

So if NRT is written in C I'd expect it to behave similarly. If so then we'd be allocating 1 byte short here as well, no?

rbc/irtools.py
```diff
@@ -36,7 +38,13 @@ def get_called_functions(library, funcname=None):
...
         for instruction in block.instructions:
             if instruction.opcode == 'call':
-                name = list(instruction.operands)[-1].name
+                instr = list(instruction.operands)[-1]
+                name = instr.name
+                # ignore in a call to a function pointer
+                if 'bitcast' in str(instr):
+                    continue
+                if name == '':
+                    continue
```
brenocfg, line 47: In the LLVM C++ API, this is usually done by checking if `CallInst->getCalledFunction() == nullptr || CallInst->getCalledFunction()->isDeclaration()`, does `llvmlite` not have similar methods?

rbc/nrt.py
```diff
@@ -0,0 +1,670 @@
...
+
+        for name in dir(self):
+            if not name.startswith("define_"):
+                continue
+
+            if name in _debug_functions and not self.debug_nrt:
+                continue
+
+            meth = getattr(self, name)
+            fn_name = name[len("define_"):]
```
brenocfg, line 337: This is some clever wizardry

rbc/nrt.py
```diff
@@ -0,0 +1,670 @@
...
+        with self.nrt_debug_ctx():
+            # Because the previously reallocated memory might be freed by
+            # realloc, RBC cannot call it in this function. Therefore, instead
+            # of reallocation, RBC allocates a new block of memory and returns
+            # it to the caller. The caller is responsible for copying memory
+            # from the old buffer into the most recent allocated one."
+            [ptr, size] = args
+            new_ptr = builder.call(self.NRT_Allocate_External, [size, NULL])
+            self.NRT_Debug("ptr=%p size=%zu -> new_ptr=%p", ptr, size, new_ptr)
+        builder.ret(new_ptr)
```
brenocfg, line 414: Is there any reason why `realloc` can't just be a wrapper around `malloc` + `memcpy` then? That would remove the weirdness about the caller being responsible for copying, which isn't the case for `realloc`. I think it's preferable to keep to the original function's API as much as possible.

rbc/heavydb/text_encoding_none.py
brenocfg: I just saw your comment later on that `allocate_varlen_buffer` allocates `len + 1` bytes, so this makes sense now :). Disregard the above comment.

brenocfg (commented):
Overall looks good, this is awesome stuff @guilhermeleobas!

I have a few questions/comments, mostly related to the API of the NRT functions. Will wait for your feedback on those before giving explicit approval.

rbc/heavydb/text_encoding_none.py
guilhermeleobas: Yes! It allocates one more byte, which we use to store the null terminator

rbc/irtools.py
guilhermeleobas: Unfortunately, no. The available methods for a ValueRef are:

```python
(Pdb++) print(instruction)
  %.32 = call i8* bitcast (%Struct.MemInfo.5* (i64, i8*)* @NRT_MemInfo_new_varsize_dtor to i8* (i64, i8*)*)(i64 64, i8* bitcast (void (i8*)* @.dtor.list.int64 to i8*))
(Pdb++) instruction.has_name
*** AttributeError: 'ValueRef' object has no attribute 'has_name'
(Pdb++) vars(instruction)
{'_kind': 'instruction', '_parents': {'module': <llvmlite.binding.module.ModuleRef object at 0x7feb8bf0bf10>, 'function': <llvmlite.binding.value.ValueRef object at 0x7feb8bf0bfa0>, 'block': <llvmlite.binding.value.ValueRef object at 0x7feb8b62e7a0>}, '_ptr': <llvmlite.binding.ffi.LP_LLVMValue object at 0x7feb8b832140>, '_as_parameter_': <llvmlite.binding.ffi.LP_LLVMValue object at 0x7feb8b832140>, '_capi': <llvmlite.binding.ffi._lib_wrapper object at 0x7fec59bedc00>}
(Pdb++) dir(instruction)
['__bool__', '__class__', '__del__', '__delattr__', '__dict__', '__dir__', '__doc__', '__enter__', '__eq__', '__exit__', '__format__', '__ge__', '__getattribute__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__le__', '__lt__', '__module__', '__ne__', '__new__', '__nonzero__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__', '__weakref__', '_as_parameter_', '_capi', '_closed', '_dispose', '_kind', '_owned', '_parents', '_ptr', 'add_function_attribute', 'arguments', 'attributes', 'block', 'blocks', 'close', 'closed', 'detach', 'function', 'instruction', 'instructions', 'is_argument', 'is_block', 'is_declaration', 'is_function', 'is_global', 'is_instruction', 'is_operand', 'linkage', 'module', 'name', 'opcode', 'operands', 'storage_class', 'type', 'visibility']
```

rbc/heavydb/text_encoding_none.py
guilhermeleobas: This function is not actually used anymore and we can remove it. But answering your question. Without the linkage as `linkonce`, we would get multiple definition errors.

brenocfg: Weird, why so? Also if it's not used anymore then I guess removing it is ok?

rbc/irtools.py
brenocfg: Bummer :(

rbc/nrt.py
brenocfg: Ahh that makes sense.

rbc/heavydb/text_encoding_none.py
guilhermeleobas: Yeah, just removed it.

rbc/heavydb/text_encoding_none.py
brenocfg: Ah yeah, it's pretty safe to assume any code that handles/creates structs directly in LLVM is intended for stack-allocated struct objects and not pointers.

guilhermeleobas:
Pinging @mflaxman10 for visibility

rbc/tests/heavydb/test_nrt_set.py
```diff
@@ -0,0 +1,103 @@
...
+    # [X] set.intersection
+    # [X] set.intersection_update
+    # [X] set.isdisjoint
+    # [X] set.issubset
+    # [X] set.issuperset
+    # [X] set.pop
+    # [X] set.remove
+    # [X] set.symmetric_difference
+    # [X] set.symmetric_difference_update
+    # [ ] set.union
```
pearu, line 36: What blocks supporting `set.union`?

rbc/unicodetype_db.h
```diff
@@ -0,0 +1,46 @@
+#include <stdint.h>
+#include "path/to/numba/_unicodetype_db.h"
+
+// To generate unicodetype_db.ll, fix the include above and compile it with
+//      clang -S -emit-llvm -O2 unicodetype_db.h -o unicodetype_db.ll
```
pearu, line 5: To avoid unnecessary changes to this file, I suggest:
```suggestion
#include "_unicodetype_db.h"

// To generate unicodetype_db.ll, fix the include above and compile it with
//      clang -S -emit-llvm -O2 -Ipath/to/numba/ unicodetype_db.h -o unicodetype_db.ll
```

rbc/irtools.py
```diff
@@ -352,6 +394,14 @@ def add_metadata_flag(main_library, **kwargs):
     main_library.add_ir_module(module)
 
 
+def read_unicodetype_db():
+    unicode_file = os.path.join(os.path.dirname(__file__),
+                                'unicodetype_db.ll')
+    with open(unicode_file, 'r') as f:
+        s = ''.join(f.readlines())
+    return llvm.parse_assembly(s)
```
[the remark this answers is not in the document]
guilhermeleobas: Done!

rbc/irtools.py
```diff
@@ -410,7 +466,7 @@ def compile_to_LLVM(functions_and_signatures,
         add_metadata_flag(main_library,
                           pass_column_arguments_by_value=0,
                           manage_memory_buffer=1)
-        main_library._optimize_final_module()
+        # main_library._optimize_final_module()
```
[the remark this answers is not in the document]
guilhermeleobas: Done!

rbc/tests/heavydb/test_nrt_set.py
guilhermeleobas: It doesn't seem to work in Numba:
```python
self = <numba.core.typeinfer.TypeInferer object at 0x7ff2e5965f30>, raise_errors = True

    def propagate(self, raise_errors=True):
        newtoken = self.get_state_token()
        oldtoken = None
        # Since the number of types are finite, the typesets will eventually
        # stop growing.

        while newtoken != oldtoken:
            self.debug.propagate_started()
            oldtoken = newtoken
            # Errors can appear when the type set is incomplete; only
            # raise them when there is no progress anymore.
            errors = self.constraints.propagate(self)
            newtoken = self.get_state_token()
            self.debug.propagate_finished()
        if errors:
            if raise_errors:
                force_lit_args = [e for e in errors
                                  if isinstance(e, ForceLiteralArg)]
                if not force_lit_args:
>                   raise errors[0]
E                   numba.core.errors.TypingError: Failed in nopython mode pipeline (step: nopython frontend)
E                   Invalid use of BoundFunction(set.union for set(int64)) with parameters (set(Literal[int](0)))
E
E                   During: resolving callee type: BoundFunction(set.union for set(int64))
E                   During: typing of call at /home/guilhermeleobas/git/rbc/rbc/tests/heavydb/test_nrt_set.py (84)
E
E
E                   File "rbc/tests/heavydb/test_nrt_set.py", line 84:
E                       def test_set(t, method):
E                           <source elided>
E                           elif method_str == 'union':
E                               return len(s.union({0})) == 6
E                               ^

../../miniconda3/envs/rbc/lib/python3.10/site-packages/numba/core/typeinfer.py:1086: TypingError
```

rbc/heavydb/text_encoding_none.py
guilhermeleobas: The code in the end is practically the same. For the sake of readability, let's keep `p1`.

pearu (approved):

PR #532, opened 2023-02-28
Fix test_daytimeinterval when server was compiled with SYSTEM_TFS=off

guilhermeleobas:
As title

### DIFF
--- a/rbc/tests/heavydb/test_daytimeinterval.py
+++ b/rbc/tests/heavydb/test_daytimeinterval.py
@@ -13,10 +13,13 @@ def heavydb():
     ):
         try:
             o.sql_execute('select * from table(generate_series(0, 10, 1));')
-        except Exception as msg:
-            pytest.skip('test requires server to be compiled with SYSTEM_TFS '
-                        f'enabled ({msg}):')
-
+        except Exception as e:
+            msg = "Undefined function call 'generate_series' in SQL statement"
+            if msg in str(e):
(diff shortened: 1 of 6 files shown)


rbc/tests/heavydb/test_daytimeinterval.py
```diff
@@ -11,6 +11,12 @@ def heavydb():
     for o in heavydb_fixture(
         globals(), debug=False, load_test_data=False, minimal_version=(6, 2)
     ):
+        try:
+            o.sql_execute('select * from table(generate_series(0, 10, 1));')
+        except Exception as msg:
+            pytest.skip('test requires server to be compiled with SYSTEM_TFS '
+                        f'enabled ({msg}):')
```
pearu, line 18: Before skipping, could you also check that `str(msg)` contains the expected error message when SYSTEM_TFS is disabled in the server? This is for sanity as try-except-skip may lead to false skips. For instance:
```python
if 'Table function called but built with ENABLE_SYSTEM_TFS=off' in str(msk):
    pytest.skip(...)
else:
    raise
```

pearu (changes requested):
Just one note but otherwise looks good.

rbc/tests/heavydb/test_daytimeinterval.py
guilhermeleobas: Changed in [d7b2d23](https://github.com/xnd-project/rbc/pull/532/commits/d7b2d23c516ee241dd33d00aa3cf4b9665c60045)

PR #538, opened 2023-04-28
Update Numba to 0.57

guilhermeleobas:
Array tests are currently broken


guilhermeleobas:
Pending https://github.com/conda-forge/numba-feedstock/pull/115

guilhermeleobas:
Blocked on HeavyDB 7.0 release on conda-forge

rbc/irtools.py
```diff
@@ -438,6 +439,27 @@ def compile_to_LLVM(functions_and_signatures,
...
+               'Please, downgrade to a previous release of Numba that uses the '
+               'same LLVM version as the HeavyDB server. For more information, '
+               'see the table below:\n\n'
+               'https://github.com/numba/llvmlite#compatibility\n'
+               'https://github.com/heavyai/heavydb#dependencies\n\n'
+               f'To disable this error, run RBC with {flag}=1 flag enabled.')
+
+        DISABLE_LLVM_MISMATCH_ERROR = int(os.environ.get(flag, False))
+        if not DISABLE_LLVM_MISMATCH_ERROR:
+            raise LLVMVersionMismatchError(msg)
```
guilhermeleobas, line 461: Would this be the most appropriated place to put this check?

pearu (changes requested):

.github/workflows/rbc_test.yml
```diff
@@ -108,11 +108,25 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        python-version: ['3.10', '3.9', '3.8']
-        numba-version: ['0.56', '0.55']
+        python-version: ['3.10', '3.9']
+        numba-version: ['0.56']
         heavydb-version: ['6.2', '6.1', '6.0']
```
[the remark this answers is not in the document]
guilhermeleobas: HeavyDB 6.4 has a bug where it crashes when executing UDFs/UDTFs.

environment.yml
```diff
@@ -14,8 +14,8 @@ dependencies:
   - setuptools
   - numba>=0.55
   - llvmlite
-  - clang==11.1
-  - clangxx==11.1
+  - clang
+  - clangxx
```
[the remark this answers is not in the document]
guilhermeleobas: Depends on the environment. On CI? Yes! In a user environment, it depends on how mamba will solve it. But `clang/clangxx` are optional dependencies.

rbc/irtools.py
guilhermeleobas: Okay, I can change that. We only check for the LLVM 11/14 combination, which is known to be problematic.

rbc/heavydb/array.py
```diff
@@ -45,12 +45,11 @@ def deepcopy(self, context, builder, val, retptr):
...
-        struct_load = builder.load(val, name='struct_load')
-        src = builder.extract_value(struct_load, 0, name='array_buff_ptr')
-        element_count = builder.extract_value(struct_load, 1, name='array_size')
-        is_null = builder.extract_value(struct_load, 2, name='array_is_null')
-
         zero, one, two = int32_t(0), int32_t(1), int32_t(2)
+        src = builder.load(builder.gep(val, [zero, zero]), name='array_buff_ptr')
+        element_count = builder.load(builder.gep(val, [zero, one]), name='array_size')
+        is_null = builder.load(builder.gep(val, [zero, two]), name='array_is_null')
+
```
pearu, line 52: Could you create a separate PR for this fix because landing the PR is blocked by heavydb 7 release.

rbc/heavydb/buffer.py
```diff
@@ -207,7 +208,10 @@ def heavydb_buffer_constructor(context, builder, sig, args):
     ptr = memalloc(context, builder, ptr_type, element_count, element_size)
 
     llty = context.get_value_type(sig.return_type.dtype)
-    st_ptr = builder.alloca(llty)
+    st_size = context.get_abi_sizeof(llty)
+    st_ptr = builder.bitcast(allocate_varlen_buffer(builder, int64_t(st_size),
+                                                    int64_t(1)),
+                             llty.as_pointer())
```
pearu, line 214: Add this to the separate PR.

rbc/heavydb/text_encoding_none.py
```diff
@@ -176,7 +176,10 @@ def text_encoding_none_unicode_ctor(context, builder, sig, args):
     uni_str = context.make_helper(builder, sig.args[0], value=args[0])
 
     llty = context.get_value_type(sig.return_type.dtype)
-    st_ptr = builder.alloca(llty)
+    st_size = context.get_abi_sizeof(llty)
+    st_ptr = builder.bitcast(allocate_varlen_buffer(builder, int64_t(st_size),
+                                                    int64_t(1)),
+                             llty.as_pointer())
     zero, one, two = i32(0), i32(1), i32(2)
```
pearu, line 183: Add this to the separate PR.

pearu (changes requested):
@guilhermeleobas since we don't know when heavydb 7 will be release (so merging this PR is blocked), could you move the `alloc` releated fixes to a separate PR that can be landed earlier?

doc/envvars.rst
```diff
@@ -14,3 +14,9 @@ Debugging
 
     If set to non-zero, insert debug statements to our implementation
     of Numba Runtime (NRT)
+
+.. envvar:: DISABLE_LLVM_MISMATCH_WARN
```
pearu, line 18:
```suggestion
.. envvar:: RBC_DISABLE_LLVM_MISMATCH_WARN
```

rbc/heavydb/remoteheavydb.py
```diff
@@ -1266,8 +1268,35 @@ def _make_udf(self, caller, orig_sig, sig):
...
+            target = self.targets['cpu']
+            server_llvm_version = target.llvm_version
+            client_llvm_version = llvm.llvm_version_info
+
+            if (server_llvm_version[0], client_llvm_version[0]) == (11, 14):
+                c_llvm = '.'.join(map(str, client_llvm_version))
+                s_llvm = '.'.join(map(str, server_llvm_version))
+                msg = (
+                    f'The client LLVM version ({c_llvm}) is greater than the server '
+                    f'LLVM version ({s_llvm}). This is known to be unsupported. '
```
pearu, line 1286: With #551, I believe this statement has become out-of-date. Could you rephrase the message along the lines of `"If you experience any issues that could be due to the discrepancy of LLVM versions, please try to downgrade/update Numba that uses LLVM version ({s_llvm})."`. Also, this statement can be generalized to all LLVM version combinations, not just `(11, 14)`.

rbc/heavydb/remoteheavydb.py
```diff
@@ -1266,8 +1268,35 @@ def _make_udf(self, caller, orig_sig, sig):
             name + sig.mangling(),
             atypes, rtype, annotations)
 
+    def _check_llvm_version(self):
+        # check LLVM version before compiling to LLVM
+        flag_name = 'RBC_DISABLE_LLVM_MISMATCH_WARN'
+        flag = int(os.environ.get(flag_name, False))
+
```
pearu, line 1275: We should check the LLVM version and possibly warn only once. One possible approach is as follows:
```suggestion
        os.environ[flag_name] = 1
```

rbc/heavydb/remoteheavydb.py
pearu: I now see that the purpose is to warn only about the problematic LLVM version combinations. Since 11 and 14 are not problematic anymore, this function is essentially a no-op? What do you think?

In fact, the combination of 14 and 11 might be more problematic as a newer client-side LLVM version may use features (such as LLVM attributes or similar) that the server LLVM IR parser cannot handle.

PR #542, opened 2023-05-05
Revise device support

### LINKED ISSUE #541: rbc tests fail when using cuda-enabled heavydb
PR #529 introduces the rbc tests failures when using cuda-enabled heavydb as reported in https://github.com/xnd-project/rbc/pull/537#issuecomment-1535011455

_Originally posted by @pearu in https://github.com/xnd-project/rbc/issues/529#issuecomment-1535066865_

pearu:
Fixes #541 .

In addition:
- fixes numerous warnings from tests when using cuda-enabled heavydb server
- auto-detect the devices support from signature types
- introduce `devices` kw argument to `external`
- deprecate CPU/GPU annotations

### DIFF
--- a/rbc/external.py
+++ b/rbc/external.py
@@ -3,8 +3,7 @@
 
 # This code has heavily inspired in the numba.extending.intrisic code
 import warnings
-from typing import Dict, List, Union
-from collections import defaultdict
+from typing import List, Union
 from numba.core import extending, funcdesc, types, typing
 from rbc.typesystem import Type
 from rbc.targetinfo import TargetInfo
@@ -15,7 +14,7 @@ from rbc.utils import validate_devices
 class External:
 
     @classmethod
-    def external(cls, *args, **kwargs):
+    def external(cls, signature, *, devices=['CPU']):
         """Parameters
         ----------
         signature : object (str, ctypes function, python callable, numba function)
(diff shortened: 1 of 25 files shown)


rbc/heavydb/row_function_manager.py
```diff
@@ -68,7 +72,7 @@ def heavydb_udf_manager_get_dict_id(mgr, func_name, arg_idx):
     if target_info.software[1][:3] < (6, 2, 0):
         raise UnsupportedError(error_msg % (".".join(map(str, target_info.software[1]))))
 
-    defn = 'int32 RowFunctionManager_getDictId(int8*, int8*, int32)'
+    defn = 'int32 RowFunctionManager_getDictId(int8*, int8*, int32)|CPU'
```
guilhermeleobas, line 75: Could we move the device specification to a keyword argument? Similar to what we do with the `@heavydb` decorator? i.e.:

```python
defn = 'int32 RowFunctionManager_getDictId(int8*, int8*, int32)'
get_dict_id = external(defn, devices=['cpu'])
```

rbc/tests/heavydb/test_text_encoding_dict.py
```diff
@@ -75,28 +75,28 @@ def test_getstringid_from_arg(x, m, text, y):
             y[0] = x.string_dict_proxy.getStringId(text)
             return 1
 
-        @heavydb('int32(TableFunctionManager, OutputColumn<TextEncodingDict> | input_id=args<>)',  # noqa: E501
-                 devices=['cpu'])
+        # Using TableFunctionManager argument implies devices=['cpu']
+        @heavydb('int32(TableFunctionManager, OutputColumn<TextEncodingDict> | input_id=args<>)')  # noqa: E501
```
guilhermeleobas, line 79: I thought that compiling this UDTF for GPU would result in a compilation failure, as the external functions used externally are declared with `|CPU` device suffix.

rbc/external.py
```diff
@@ -79,7 +77,16 @@ def name(self):
...
+        target_info = TargetInfo()
+        device = "CPU" if target_info.is_cpu else "GPU"
+        if device not in self._signatures:
+            satypes = ", ".join(map(str, atypes))
+            compile_target = target_info.get_compile_target()
+            raise TypeError(
+                f"{compile_target=}: no signatures with argument types `{satypes}` and"
+                f" device `{device}` while processing `{self.name}`"
+                " (perhaps the compile target cannot support the device?)"
+            )
```
guilhermeleobas, line 89: Isn't the `TypeError` below doing the same thing? Raise an exception if no function for the given signature and device is found.

rbc/heavydb/row_function_manager.py
pearu: I think this makes sense. We should also get rid of the `| CPU` annotation as it duplicates the `devices` kw argument. What do you think?

rbc/tests/heavydb/test_text_encoding_dict.py
pearu: It failed badly in `External::match_signature` because the external functions were not available for the `GPU` target while now the GPU target will be skipped in `match_signature` so that the compilation for `CPU` could finish.

rbc/external.py
pearu: Here, the type error is raised because a signature with the target device is missing while below the error is raised because a signature with the given argument types is missing.

The first case happens when, say, when target info is GPU while a signature has `TableFunctionManager` argument that is available only for the CPU target (hence, there is no `GPU` device key in `self._signatures`).

rbc/heavydb/row_function_manager.py
guilhermeleobas: Agreed. It can be done in a follow-up PR.

pearu: Done: introduced `devices` kw argument to `external` and deprecated CPU/GPU annotations.

rbc/external.py
pearu: I have revised this code. So, please review the update code.

rbc/external.py
```diff
@@ -2,69 +2,96 @@
...
+from rbc.utils import validate_devices
 
 
 class External:
+
     @classmethod
-    def external(cls, *args):
-        """
-        Parameters
+    def external(cls, *args, **kwargs):
```
guilhermeleobas, line 18: Given that all use cases we have for this function is a single argument, what about replacing it by:

```suggestion
    def external(cls, signature, *, devices=["CPU"]):
```

pearu: Done. This also simplified external implementation.

rbc/external.py
```diff
@@ -2,106 +2,125 @@
...
+        if not name:
+            raise ValueError(
+                f"external function name not specified for signature {signature}"
+            )
+
+        # Default devices contains `CPU` only because in most cases
+        # the externals are defined for the CPU target.
+        devices = validate_devices(devices)
+        # Using annotation devices (specified via `| CPU` or `| GPU`) is deprecated.
+        # TODO: turn the warning to an exception for rbc version 0.12 or higher
```
guilhermeleobas, line 43: Could you create an issue for that?

pearu: Done: https://github.com/xnd-project/rbc/issues/544

PR #545, opened 2023-05-09
Fix readthedocs CI failure

pearu:
As in the title


guilhermeleobas:
Correct me if I'm wrong. Setuptools is built on top of distutils, which is being deprecated in Python 3.12. Perhaps it is better if we switch to more modern build tool and replace existing `setup.cfg/setup.py` by `pyproject.toml`.

guilhermeleobas (approved):

PR #549, opened 2023-05-16
Use new external keyword arg

guilhermeleobas:
This is a follow-up PR of #542, which adapts existing usages of `external` to use the new keyword argument `devices`.


pearu (approved):

Issue #548, opened 2023-05-15
Add support for Column<TextEncodingNone>


guilhermeleobas:
As title

PR #550, opened 2023-05-16
Simplify ext_arguments_map ctor

guilhermeleobas:
part of #548


rbc/tests/heavydb/test_type_to_extarg.py
```diff
@@ -0,0 +1,148 @@
...
+    "float32",
+    "float64",
+    "int8*",
+    "int16*",
+    "int32*",
+    "int64*",
+    "float32*",
+    "float64*",
+    "bool",
+    "bool*",
```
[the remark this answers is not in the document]
pearu: Also, add `double` and `float`.

rbc/tests/heavydb/test_type_to_extarg.py
```diff
@@ -0,0 +1,148 @@
...
+    "DayTimeInterval",
+    "YearMonthTimeInterval",
+    # Array
+    "Array<bool>",
+    "Array<int8_t>",
+    "Array<int16_t>",
+    "Array<int32_t>",
+    "Array<int64_t>",
+    "Array<float>",
+    "Array<double>",
```
pearu, line 50: Also add `Array<int8>`, ..., `Array<float32>`, `Array<float32_t>` etc.

rbc/tests/heavydb/test_type_to_extarg.py
```diff
@@ -0,0 +1,148 @@
...
+    "Array<GeoPolygon>",
+    "Array<GeoMultiPolygon>",
+    # Column
+    "Column<bool>",
+    "Column<int8_t>",
+    "Column<int16_t>",
+    "Column<int32_t>",
+    "Column<int64_t>",
+    "Column<float>",
+    "Column<double>",
```
pearu, line 66: Same: add `Column<int8>`, etc.

rbc/tests/heavydb/test_type_to_extarg.py
```diff
@@ -0,0 +1,148 @@
...
+    "Column<Timestamp>",
+    "Column<GeoPoint>",
+    "Column<GeoMultiPoint>",
+    "Column<GeoLineString>",
+    "Column<GeoMultiLineString>",
+    "Column<GeoPolygon>",
+    "Column<GeoMultiPolygon>",
+    # ColumnList
+    "ColumnList<bool>",
+    "ColumnList<int8_t>",
```
pearu, line 78: Same: add `ColumnList<int8>`, etc

rbc/tests/heavydb/test_type_to_extarg.py
```diff
@@ -0,0 +1,148 @@
...
+    "ColumnList<TextEncodingNone>",
+    "ColumnList<GeoPoint>",
+    "ColumnList<GeoMultiPoint>",
+    "ColumnList<GeoLineString>",
+    "ColumnList<GeoMultiLineString>",
+    "ColumnList<GeoPolygon>",
+    "ColumnList<GeoMultiPolygon>",
+    # ColumnArray
+    "Column<Array<bool>>",
+    "Column<Array<int8_t>>",
```
pearu, line 94: Same as above.

rbc/tests/heavydb/test_type_to_extarg.py
```diff
@@ -0,0 +1,148 @@
...
+    "Column<Array<TextEncodingNone>>",
+    "Column<Array<GeoPoint>>",
+    "Column<Array<GeoMultiPoint>>",
+    "Column<Array<GeoLineString>>",
+    "Column<Array<GeoMultiLineString>>",
+    "Column<Array<GeoPolygon>>",
+    "Column<Array<GeoMultiPolygon>>",
+    # ColumnListArray
+    "ColumnList<Array<bool>>",
+    "ColumnList<Array<int8_t>>",
```
pearu, line 110: Same as above.

rbc/heavydb/remoteheavydb.py
```diff
@@ -929,85 +929,33 @@ def _get_ext_arguments_map(self):
 
         for T, Tname in [
                 ('bool', 'Bool'),
-                ('int8_t', 'Int8'),
-                ('int16_t', 'Int16'),
-                ('int32_t', 'Int32'),
-                ('int64_t', 'Int64'),
-                ('float', 'Float'),
-                ('double', 'Double'),
```
pearu, line 937: I think we'll need these.

rbc/tests/heavydb/test_type_to_extarg.py
```diff
@@ -0,0 +1,148 @@
...
+    "ColumnList<Array<GeoMultiLineString>>",
+    "ColumnList<Array<GeoPolygon>>",
+    "ColumnList<Array<GeoMultiPolygon>>",
+)
+
+
+def test_len(heavydb):
+    msg = 'variable "all_types" requires update'
+    assert len(heavydb._get_ext_arguments_map()) == len(all_types), msg
+
```
pearu, line 130: I wonder, should we also test the exceptions from user-specified ill-formed types names? For example, when they specify `Array<complex>` in the heavydb signature, will the error message be informative enough for users to not struggle with the unsupported type?

pearu (changes requested):
Overall looks good! Thanks, @guilhermeleobas!

I have a few suggestions to extend the set of supported types.

rbc/tests/heavydb/test_type_to_extarg.py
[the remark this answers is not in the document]
guilhermeleobas: Done

rbc/tests/heavydb/test_type_to_extarg.py
guilhermeleobas: Done

rbc/tests/heavydb/test_type_to_extarg.py
guilhermeleobas: Done

rbc/tests/heavydb/test_type_to_extarg.py
guilhermeleobas: Done

rbc/tests/heavydb/test_type_to_extarg.py
guilhermeleobas: Done

rbc/tests/heavydb/test_column_basic.py
```diff
@@ -902,6 +902,17 @@ def test_column_array_rewire(heavydb, typ, kind, inner):
             assert type(col) is expected
 
 
+@pytest.mark.xfail
```
pearu, line 905: Why xfail?

rbc/heavydb/remoteheavydb.py
pearu: I think we still need these. What do you think?

pearu (commented):
I have two questions. Otherwise looks good to me.

guilhermeleobas:
> I think we still need these. What do you think?

Could you think of any example that these types are required?

pearu:
> > I think we still need these. What do you think?
> 
> Could you think of any example that these types are required?

These are not required per se. I am just thinking about C/C++ users who might write, say, `Array<double>` or `Column<int32_t>`.

Also, these types are included in [test_type_to_extarg.py](https://github.com/xnd-project/rbc/pull/550/files#diff-f614180c1c47fb12101ea9de5d68b9f5070425d5f02a0f050db025f618e2eb20) as primitives.

Another approach is to have a unique list of types that rbc supports and reject all their equivalent variations with an informative exception message such as `"Array<double> is not supported, use Array<float64> instead"`.

rbc/heavydb/remoteheavydb.py
```diff
@@ -925,88 +926,47 @@ def _get_ext_arguments_map(self):
...
+                ('int8', 'Int8'),
+                ('int16', 'Int16'),
+                ('int32', 'Int32'),
+                ('int64', 'Int64'),
+                ('float32', 'Float'),
+                ('float64', 'Double'),
```
guilhermeleobas, line 936: Check to see if this part is needed?

PR #551, opened 2023-05-16
Fix heavydb server (built with llvm 14) crash when using numba with llvm 11

pearu:
As in the title.


guilhermeleobas (approved):

guilhermeleobas:
Thanks @pearu

PR #552, opened 2023-05-17
Cache the NRT module and introduce rbc.config to keep track of env vars

guilhermeleobas:
As title. This PR enables caching the NRT module, instead of regenerating it each time an UD[T]F is compiled. It also creates a `rbc/config.py` file to manage environment variables.

### DIFF
--- a/rbc/irtools.py
+++ b/rbc/irtools.py
@@ -1,8 +1,6 @@
 # Author: Pearu Peterson
 # Created: February 2019
 
-import functools
-import os
 import re
 import warnings
 from collections import defaultdict
@@ -16,7 +14,8 @@ from numba.core import errors as nb_errors
 from numba.core import extending, imputils, registry, sigutils, typing
 
 from rbc.externals import stdio
-from rbc.nrt import create_nrt_functions
+from rbc.nrt import create_nrt_functions, read_unicodetype_db
+from rbc import config
 
 from .errors import UnsupportedError
 from .libfuncs import Library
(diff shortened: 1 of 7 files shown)


rbc/irtools.py
```diff
@@ -475,6 +467,15 @@ def compile_to_LLVM(functions_and_signatures,
                     succesful_fids.append(fid)
                     function_names.append(fname)
 
+        if nrt_required(main_module):
+            if config.ENABLE_NRT:
+                main_module.link_in(unicodetype_db)
+                main_library.add_ir_module(nrt_module)
+            else:
+                msg = ("NRT required but not enabled. Set 'RBC_ENABLE_NRT=1' "
+                       "to fix this warning.")
```
pearu, line 476:
```suggestion
                msg = ("NRT is disabled but required. Undefine the environment variable 'RBC_ENABLE_NRT' or set its value to '1'")
```

rbc/tests/heavydb/test_nrt.py
```diff
@@ -0,0 +1,22 @@
...
+def test_nrt_warning(heavydb):
+
+    @heavydb("i32(i32)", devices=['cpu'])
+    def list_append(t):
+        lst = list()
+        lst.append('abc')
+        return len(lst)
+
+    with override_config('ENABLE_NRT', 0):
+        with pytest.warns(UserWarning, match='NRT required but not enabled'):
```
pearu, line 21: Update the match argument according to the suggestion above.

pearu (changes requested):
I have one nit, otherwise looks good! I like the rbc/config.py approach.
Thanks, @guilhermeleobas!

pearu (changes requested):
I have one nit, otherwise looks good! I like the rbc/config.py approach.
Thanks, @guilhermeleobas!

PR #572, opened 2023-06-30
Change methods to receive a python native str

guilhermeleobas:
This change HeavyDB methods to receive a python string rather than a TextEncodingNone


pearu (approved):
LGTM! Thank, @guilhermeleobas!

Before landing, this PR requires another iteration to resolve conflicts.

